Issue #8, opened 2017-02-20
roxma/nvim-completion-manager source?


autozimu:
As deoplete does not support async source, right now completion function is implemented as sync calls.

Might even better support async completion.

roxma/nvim-completion-manager looks cool and actively maintained/developed.

@roxma

Shougo:
> As deoplete does not support async source, right now completion function is implemented as sync calls.

Of course it is on the todo list.
But I must refactor the implementation to support the feature.
And I have denite.nvim and neovim supports...

autozimu:
@Shougo Thanks for the info.

I know you might be busy with other stuff, but could you given a rough estimation of when the work will start?

You can let me know if you need some more hand on that.

Shougo:
> I know you might be busy with other stuff, but could you give a rough estimation of when the work will start?

https://github.com/Shougo/deoplete.nvim/issues/407

Hm.

Current loadmap:

* Split the gather candidates and the completion popup
* Add the async support
* Change option API like denite.nvim
* completefunc support?
* Parallel support?
* Vim8 support?

The async support is not so far, but it depends on the other works.

autozimu:
Great. Thanks for sharing.

Shougo:
I will start the implementation ASAP.

Shougo:
> As deoplete does not support async source, right now completion function is implemented as sync calls.

I have implemented the feature.

autozimu:
Cool! I will start upgrading the integration.

Shougo:
Thanks.
I will test it.

PR #10, opened 2017-02-24
integration with nvim-completion-manager, use callback style.  #8

roxma:
Tested on my machine using [PHP language server](https://github.com/felixfbecker/php-language-server)

```vim
    \ 'php': ['php', 'path/to/language/server/base/vendor/felixfbecker/language-server/bin/php-language-server.php'],
    \ }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ More recordings at <https://github.com/autozimu/images/tree/master/LanguageClien
 # Features
 
 - Non-blocking asynchronous calls.
-- [Sensible completion](https://github.com/autozimu/images/tree/master/LanguageClient-neovim#completion). Integrated with [deoplete](https://github.com/Shougo/deoplete.nvim).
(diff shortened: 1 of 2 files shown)


roxma:
related issue #8

roxma:
There're some considerations I've omitted here:

1. If the completion calculation is slow, tha results will be ignored if the user types more characters. This doest not affect the expected behavior, but if it's very slow, and the user types fast, there may be several outdated request still processing by lanugage-server, and this might affect performance.
2. I'm not sure the way I get the `startcol` for completion is right, It seems the language server does not return this field.

autozimu:
Overall looks good to me. (I will handle small style issues after merge it in.)

As for your concerns
1. According to protocol specification, it is possible to cancel previously initiated but unfinished calls. Could be revisited later.
2. As far as I know, the server does not return startcol.

Thanks for your work!

PR #17, opened 2017-03-03
Fix #15, and recognize triggerCharacters

### LINKED ISSUE #15: PHP language server issues
I've come across a couple more issues while testing compatibility with the php language server:

1. Using nvim-completion-manager, completion shows `null` in the list instead of the correct variable name, in `completionManager_refresh` doing the following gives me a better menu:
```
-                e['abbr'] = item.get('insertText', "")
+                if 'insertText' in item and item['insertText']:
+                    e['abbr'] = item.get('insertText', "")
```

Though when I select an item from the popup menu, it repeats the `$`, so if I type `$` and then select `$foo` from the menu, I get `$$foo`

2. in `showDiagnosticMessage()`, `entry["code"]` is an int, not a string, so `msg += code` fails, `msg += str(code)` works

3. In `LanguageClientStart()`, because `textDocument_didOpen` is being called straight after `self.initialize([])`, the `textDocument_didOpen` request is being sent before the InitializeResponse is received. The [specification](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize) says:

> Until the server has responded to the initialize request with an InitializeResult the client must not sent any additional requests or notifications to the server.

This is causing completion to not work until I open a file after php-language-server has initialized.

Again, thanks for your work, this is really cool

roxma:
This PR should fix #15 , along with https://github.com/roxma/nvim-completion-manager/commit/a956998daae6a4a0617abafd99aca1792d166069

Thanks to @prabirshrestha , now It can recognize triggerCharacters returned from language server.

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -196,6 +196,8 @@ class LanguageClient:
                 priority=9,
                 scopes=[languageId],
                 abbreviation='',
+                # general pointer patterns
+                cm_refresh_patterns=[r'\.$',r'->$',r':'],
                 cm_refresh='LanguageClient_completionManager_refresh'))
         except Exception as ex:
             logger.warn("register completion manager source failed.")
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -196,8 +196,6 @@ class LanguageClient:
                 priority=9,
                 scopes=[languageId],
                 abbreviation='',
-                # general pointer patterns
-                cm_refresh_patterns=[r'\.$',r'->$',r':'],
                 cm_refresh='LanguageClient_completionManager_refresh'))
         except Exception as ex:
             logger.warn("register completion manager source failed.")
@@ -669,16 +667,13 @@ call fzf#run(fzf#wrap({{
         info = args[0]
         ctx = args[1]
 
-        typed = ctx["typed"]
+        if ctx["typed"] == "":
+            return
         col = ctx["col"]
-        kwtyped = re.search(r'[0-9a-zA-Z_]*?$', typed).group(0)
-        startcol = col-len(kwtyped)
 
         args = {}
         args["line"] = ctx["lnum"] - 1
         args["character"] = ctx["col"] - 1
-        if typed == "":
-            return
 
         uri, line, character = self.getArgs(
                 [args], ["uri", "line", "character"])
@@ -708,7 +703,7 @@ call fzf#run(fzf#wrap({{
                 matches.append(e)
 
             self.nvim.call('cm#complete', info['name'], ctx,
-                           startcol, matches, isIncomplete, async=True)
+                           ctx['startcol'], matches, isIncomplete, async=True)
 
         # Make sure the changing is synced.  Since `TextChangedI` will not be
         # triggered when popup menu is visible and neovim python client use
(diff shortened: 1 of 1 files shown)


rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -190,19 +190,46 @@ def start(self) -> None:
...
-                scopes=[languageId],
-                abbreviation='',
-                cm_refresh='LanguageClient_completionManager_refresh'))
-        except Exception as ex:
-            logger.warn("register completion manager source failed.")
-
         logger.info('End LanguageClientStart')
 
-        self.initialize([])
+        def cb(result: Dict):
```
autozimu: Can you make this a separate function and call it from self.handleInitializeResponse using self.nvim.async_call?

rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -190,19 +190,46 @@ def start(self) -> None:
...
+
+            # Note: place the callback here to capture the languageId, which is
+            # the `filetype`. Otherwise it will fail if the user change the
+            # buffer before this callback is called.
+
+            trigger_patterns = []
+            try:
+                triggerCharacters = []
```
autozimu: Is this line redundant?

rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -190,19 +190,46 @@ def start(self) -> None:
...
+                    cm_refresh_patterns=trigger_patterns,
+                    abbreviation='',
+                    cm_refresh='LanguageClient_completionManager_refresh'))
+                logger.info("register completion manager source ok.")
+            except Exception as ex:
+                logger.exception("register completion manager source failed.")
+
+            self.handleInitializeResponse(result)
+
+        # Note: If I don't use async_call here, I'll get exception when
```
autozimu: Please remove this comment.

This is a known neovim python-client limitation. It will error out when calling neovim functions if not on main thread.

rplugin/python3/LanguageClient/LanguageClient.py
roxma: This is intentional, 
```
Note: place the callback here to capture the languageId, which is
the `filetype`. Otherwise it will fail if the user change the
buffer before this callback is called.
```

If I place it in other functions, then I have to get the languageId again, which may not be the same languageId if the user switch buffer befure that.

autozimu:
And one more thing, your commits are not associated with your email account properly.

Check your gitconfig if your would like get it resolved.

autozimu:
You can create a variable 'self.languageId', as right now only one instance
of language server is supported at one time.

roxma:
Just curious, why not wrap all callback with `async_call` so that it will be executing in the main thread?

autozimu:
Good question.

My considerations are
- Main thread is responsible to handling all autocmds, like 'TextChanged' and 'CursorMoved' etc, especially considering deoplete source is running solely on main thread right now. The main thread will have enough resource to handle all these events in time if we put all possible background tasks onto secondary thread.
- Main thread is shared across all python plugins, same motivation as above.

roxma:
@autozimu I've move the nested function into handleInitializeResponse

autozimu:
Cool, thanks a lot!

I will merge it when I'm back online.

roxma:
By the way, is there any plan for multi server support? I should be useful for php developers editing javascript file.

autozimu:
Good call.

Will work on that when I have free weekend.

autozimu:
Multiple instance of language server is here now! https://github.com/autozimu/LanguageClient-neovim/commit/cff33d1b183e591f3d73029ddf032ed23e10a940

PR #42, opened 2017-04-18
Fix #41: Use rustup to run rls

### LINKED ISSUE #41: Incorrect rls command in example settings
When installed through rustup the rls binary isn't actually added to ~/.cargo/bin/rls, I'm guessing there's one leftover on your system from building it locally and using cargo install which can be removed with 'cargo uninstall rls' .

The recommended way to run it now is 'rustup run nightly rls' b/c the binary is nested in one of rustup's toolchain folders which will differ by platform.

I'll submit a PR with a fix shortly.

chetgurevitch:
ref #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Example configuration
 set hidden
 
 let g:LanguageClient_serverCommands = {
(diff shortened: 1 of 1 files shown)


autozimu:
Great. Thank you.

PR #44, opened 2017-04-24
Log the Exception preventing completion from registering

oxinabox:
If something has failed,
it is nice to know as much as possible about why it has failed.

So lets log that exception for why the completion manager can't register.


autozimu:
Thanks.

mqudsi:
All this prints for me is `register completion manager source failed. Error: NvimError(b'Error calling function.',)`

I wish it would say _which_ function :/

autozimu:
The message is because this plugin will always try register itself as a source for <https://github.com/roxma/nvim-completion-manager>. It is expected to fail when you don't have NCM installed, thus can safely ignore if you don't use NCM.

PR #49, opened 2017-05-20
Insert use function

dantleech:
This PR adds a "insert use" function and can be used to add a use statement for the word (class short name) under the cursor (selecting from FZF):

```
nnoremap <silent><Leader>u :call LanguageClient_insert_use({"name": expand('<cword>')})<CR>
```

This is pretty crude (i'm not a Python developer and only looked at nvim today) and not sure if this is a valid feature for this library, or indeed if there is a better way to do this already.

So feel free to close if it's inappropriate, otherwise willing to do more work on it.

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -63,6 +63,7 @@ class LanguageClient:
         self.diagnostics = {}
         self.lastLine = -1
         self.hlsid = None
+        self.insertUseName = ""
         self.signs = []
         self.serverCommands = {}
         self.changeThreshold = 0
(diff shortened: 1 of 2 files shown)


rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -63,6 +63,7 @@ def __init__(self, nvim):
         self.diagnostics = {}
         self.lastLine = -1
         self.hlsid = None
+        self.insertUseName = ""
```
dantleech: The callback needed to know the name of the class in order to filter the results - so had to add this hack - don't know if there is a way to avoid this.

dantleech:
Hmm, this is PHP specific :P

PR #53, opened 2017-06-03
Shorten rls command in readme

chetgurevitch:
ref #41


autozimu:
The user might be using versions other than nightly as default, in which case 'rustup run nightly' is necessary.

The changes would make sense in personal configuration, but for the readme file, I'd like it to cover as most use cases as possible.

chetgurevitch:
Ah true, forgot about that

PR #56, opened 2017-06-07
keep deoplete and nvim completion consistent

HerringtonDarkholme:
Keep deoplete the same as  `convert_lsp_completion_item_to_vim_style`


autozimu:
Thanks

PR #58, opened 2017-06-11
Use tempfile lib for log file

evolarium:
The fixed name for the log file causes issues with multiple users.  The
tempfile library will create temporary files that don't conflict.


autozimu:
Thanks.

PR #71, opened 2017-07-08
Include instruction to install Python typing module

scooter-dangle:
I don't know enough about Python environments to be able to say if this is a universal requirement, but on OS X, I had to install `typing` to get `:LanguageClientStart` to succeed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,11 +2,19 @@
 
 Obviously you need [neovim](https://github.com/neovim/neovim#install-from-package)!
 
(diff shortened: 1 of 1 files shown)


autozimu:
`typing` module is included in python3 >= `3.6`.

So please insert a note indicating it is only needed for python3 < `3.6`. Thanks.

scooter-dangle:
> So please insert a note indicating it is only needed for python3 < 3.6.

✅

autozimu:
Thanks.

PR #81, opened 2017-07-30
Fix #67: show line content in locations list

### LINKED ISSUE #67: LanguageClient_textDocument_references -- add line
I use the **LanguageClient** with the python server, and don't really know on which side is responsible. But I think the `LanguageClient_textDocument_references` function should not only populate [Denite](https://github.com/Shougo/denite.nvim) with the file and line information, but also show the content of the line.

KonanMentor:
Fixes #67

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -4,6 +4,7 @@ import subprocess
 import json
 import inspect
 import threading
+import linecache
 from functools import partial
 from typing import List, Dict, Union, Any  # noqa: F401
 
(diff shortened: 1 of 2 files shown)


autozimu:
Could you please check the tests failure?

KonanMentor:
@autozimu, unfortunately I am not able to reproduce this failure locally.

```sh
docker build . -t autozimu/languageclientneovim
docker run -v $(pwd):/root/.config/nvim autozimu/languageclientneovim bash -c "cd /root/.config/nvim/rplugin/python3 && make"
...
========================== 25 passed in 54.84 seconds ==========================
```

Could you please assist with it? Does it fail on your machine?

autozimu:
Sure. I will try run it locally and put up the results here.

autozimu:
It failed on my local as well.

```
assert Counter() == Counter({'fn abcgreet() -> i32 {\n': 1})
```
This line shall always fail from my point of view.

Maybe you were running tests based on master branch instead of your own feature branch?

KonanMentor:
Oh, it seems that I've found a clue. The test passes with freshly-built docker image, but it fails with the `autozimu/languageclientneovim` image from Docker Hub, which was built 3 months ago. Probably some dependencies have been changed since that time. (I suspect rust language server, but I'm not sure)

So there is at least one solution: rebuild docker image with updated dependencies and push it to Docker Hub. I could try to determine the dependency that causes this test failure if it's necessary though.

autozimu:
You could just run tests locally if you are running Linux or macOS.

KonanMentor:
Okay, so my assumptions were correct. The test fails because of outdated `rls`
```
$ docker run -it -v $(pwd):/root/.config/nvim autozimu/languageclientneovim /bin/bash
# cd /root/.config/nvim/rplugin/python3/
# make
==== 1 failed, 9 passed in 42.92 seconds
# /root/.cargo/bin/rustup update nightly
# make
==== 25 passed in 54.88 seconds
```

Are you ok with rebuilding the image that is used for `travis-ci` testing, so new version of rust toolchain will be used?

autozimu:
I've updated the docker image, as you might don't have permission to override it.

Tests passed. Merging in. Thank you.

PR #90, opened 2017-08-11
add support for haskell projects

gleachkr:
This finds the root of haskell projects, and results in partial support for the haskell-ide-engine


autozimu:
Thanks.

PR #98, opened 2017-08-16
Remove last param from setqflist and setloclist because it's unused

DeltaEvo:
I'm not sure if it's the same with the python binding
According to vim documentation last param for setqflist and setloclist is a Dictionary so current value is invalid

http://vimdoc.sourceforge.net/htmldoc/eval.html#setqflist()

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -973,11 +973,9 @@ call fzf#run(fzf#wrap({{
         self.asyncCommand(cmd)
 
         if self.diagnosticsList == "quickfix":
-            self.nvim.funcs.setqflist(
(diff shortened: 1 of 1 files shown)


autozimu:
Neovim version do support a third param, which should be a dict.

I was trying to set quicklist title.

Would you mind update this PR by changing the last param to `{"title": "LanguageClient-diagnostics"}` instead of removing it?

DeltaEvo:
I don't even checked that the doc of vim is the same as neovim, but the doc say
```
If the optional {what} dictionary argument is supplied, then                                                                                                                                                                 
                only the items listed in {what} are set. The first {list}                                                                                                                                                                    
                argument is ignored.
```
So if i add this param the quicklist will not be filled

autozimu:
I misunderstanded the docs. Thanks.

PR #103, opened 2017-08-20
Adds User autocmds for started and stopped

BoltsJ:
Adds two autocmds LanguageClientStarted and LanguageClientStopped that trigger after LanguageClient is started and stopped respectively.


autozimu:
Thanks.

BoltsJ:
No problem!

Issue #107, opened 2017-08-22
Goto and Rename not working with RLS


chrisduerr:
- neovim version: 0.2.0
- Plugin version: master (5781243)
- Minimal vimrc
- Content of `:CheckHealth`: https://pastebin.com/At1UEcgF
- Content of `~/.local/share/nvim/rplugin.vim`: https://pastebin.com/tJi2zV0j
- content of `/tmp/LanguageClient.log`: https://pastebin.com/YBsU6e0a

So I have recently reinstalled my computer and ever since I haven't been able to get much useful stuff out of RLS. I went over the install guide a few times but I see nothing I could have missed. While syntax errors work sometimes, stuff like rename or goto never do. Until now I have just ignored it thinking that it might just be buggy right now but it has been bothering me a lot.

Any idea what I could try to make this work? Chances are the mistake is somewhere on my end, but I'd be glad if someone would still be able to help me out.

DeltaEvo:
Can you change log level  ([help](https://github.com/autozimu/LanguageClient-neovim/blob/master/doc/LanguageClient.txt#L215)) and resend the content of `/tmp/LanguageClient.log` please ?

chrisduerr:
That first part is the normal setting and at the bottom it's with the debug log level. There is info and debug. Is there any other log level not mentioned in the help?

DeltaEvo:
Oh yes i'm sorry but some parts are missing. Can you change log level before starting the language server ?

autozimu:
The log is not complete.

Please follow steps to get a new complete log file
- cleanup /tmp/LanguageClient.log
- reopen neovim
- set LanguageClient logging level
- resend all content of /tmp/LanguageClient.log

I would also recommend you try a simple project, like https://github.com/autozimu/LanguageClient-neovim/tree/master/rplugin/python3/tests/sample-rs to see if it works.

Ronaldho80:
I have a similar Problem. Nothing works. Only the autocompletion works. But not as it should, I guess. I set LanguageClientLoggingLevel to 'DEBUG'. This is the output:
```
$ cat /tmp/LanguageClient.log 
21:02:27 WARNING  no handler implemented for rustDocument_diagnosticsBegin
21:02:27 WARNING  register completion manager source failed. Error: NvimError(b'Error calling function.',)
21:02:30 WARNING  no handler implemented for rustDocument_diagnosticsEnd
```

DeltaEvo:
[...]

PR #108, opened 2017-08-23
Make env configurable

DeltaEvo:
This could simplify configuration (see #107)

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -20,9 +20,18 @@ Before using of LanguageClient, it is necessary to specify commands that are
 going to be used to start language server. See |LanguageClient_serverCommands|
 for detail. Here is a simple example config: >
 
-    let g:LanguageClient_serverCommands = {
-        \ 'rust': ['rustup', 'run', 'nightly', 'rls'],
-        \ }
+ let g:LanguageClient_serverCommands = {
+    \ 'rust': {
+    \   'command': ['rustup', 'run', 'nightly', 'rls'],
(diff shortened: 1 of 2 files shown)


autozimu:
Instead of specifying a different env, what about of passing current env? Which release the burdens of user from configuration once more.

DeltaEvo:
We need user env for some variables like PATH (and some language clients could use others ?), so I don't know what to do ... :smile:

autozimu:
I mean in most cases, user would still need to configure those environment variables in their shell rc files to make language servers work. nvim will be started from that same shell environment, so will nvim python host process, so will the langauge server process this plugin launch. Going this way, their language servers will not only work for this plugin, but also VSCode, Emacs etc.

Whenever language server needs access to extra environment variables, user shall went to modify their shell rc files instead of this plugin settings.

Thanks for the work whereas I don't see significant benefits by adding this extra layers within this plugin.

DeltaEvo:
rls-vscode set https://github.com/rust-lang-nursery/rls-vscode/blob/5729f5df951b7f932e7d1f6cd47dcdf0083186c1/src/extension.ts#L43 RUST_SRC_PATH if it's not set

I made this pr after #107 because this is more easy to have a one copy paste to set up a language server

I understand your point of view, i export my env variables in my shell rc too, but I think having all the conf to run the language server is more noob firendly

autozimu:
In terms of user friendly, I believe the right approach would be replicating the same logic from rls-vscode. Adding extra layer can only create more confusion.

PR #116, opened 2017-08-28
[RFC] Automatic opening of quickfix when there messages change and closing when there's not messages

purpleP:
@autozimu This is a work in progress. Notice how I used `InsertLeave autocommand`. I wanted an autocmd that would trigger just once, but I don't know how to do that with nvim api. Basically I wanted this
```
augroup OpenErrors
    au!
    au InsertLeave * copen ... | silent au! OpenErrors
augroup END
```
The part where an autocmd group is removed is probably should be in a separate function, I don't remember exactly.
Do you know how to do that with nvim?

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -71,7 +71,8 @@ class LanguageClient:
         self.changeThreshold = 0
         self.trace = "off"  # trace settings passed to server
         self.languageServerLogFilePath = os.path.join(
-                os.getenv("TMP", "/tmp"), "LanguageServer.log")
+            os.getenv("TMP", "/tmp"), "LanguageServer.log"
+        )
         self.autoStart = self.nvim.vars.get(
             "LanguageClient_autoStart", False)
         self.open = False
@@ -251,9 +252,10 @@ class LanguageClient:
                 self.selectionUI = "fzf"
             else:
                 self.selectionUI = "location-list"
-        
+
         if self.diagnosticsList == 'quickfix':
             self.get_errorlist = self.nvim.funcs.getqflist
+
             def set_errorlist(errors):
                 self.nvim.funcs.setqflist(errors, 'r')
             self.set_errorlist = set_errorlist
@@ -264,9 +266,10 @@ class LanguageClient:
             self.get_errorlist = self.nvim.funcs.getloclist
             self.set_errorlist = set_errorlist
             close_cmd, open_cmd = 'lclose', 'lopen'
-        
+
         def open_errorlist(height=None):
-            self.nvim.command((open_cmd + ' ' + str(height)) if height else open_cmd)
+            cmd = (open_cmd + ' ' + str(height)) if height else open_cmd
+            self.nvim.command(cmd)
 
         def close_errorlist():
             self.nvim.command(close_cmd)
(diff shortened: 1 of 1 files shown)


purpleP:
@autozimu Seems like It's working, would you please review the changes?

autozimu:
I won't review it until until travis pass.

purpleP:
@autozimu Do you have any idea why it's failing? I see the change that's producing the error, but not the error itself. Also how to run tests? `pytest` fails in every `LanguageServer_test.py` which doesn't seem right.

autozimu:
You care probably calling a sync function inside an async context. neovim-python doesn't support that.

Please refer to other places of the code if you want to communicate with neovim.

To run the tests, cd into python3 folder, run `make`.

purpleP:
@autozimu Last time I tried there were only style errors. Not all of them which were mine, I fixed some of them, but I'm not sure that I should have.
Also make gives me error code but doesn't show what was the problem.

purpleP:
@autozimu Notice that I used check for `buftype` in in `CursorMoved` autocmd. It's triggered in any filetype, while It only makes sense to call this function only when LanguageServer have started and only in filetypes that are in use config.

rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -70,9 +71,14 @@ def __init__(self, nvim):
         self.changeThreshold = 0
         self.trace = "off"  # trace settings passed to server
         self.languageServerLogFilePath = os.path.join(
-                os.getenv("TMP", "/tmp"), "LanguageServer.log")
+            os.getenv("TMP", "/tmp"), "LanguageServer.log"
+        )
         self.autoStart = self.nvim.vars.get(
             "LanguageClient_autoStart", False)
+        self.open = False
```
autozimu: Please come up with meaningful names.

rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -70,9 +71,14 @@ def __init__(self, nvim):
...
+        self.diagnosticsList = None
+        self.get_errorlist = None
```
autozimu: Please define these two as regular methods instead of ad-hoc variables.

rplugin/python3/LanguageClient/LanguageClient.py
```diff
@@ -983,14 +1013,33 @@ def addHighlightAndSign(self, params):
...
+        if messages_nr:
+            self.open_errorlist(messages_nr)
+        else:
+            self.close_errorlist()
+        if prev_winnr != self.nvim.funcs.winnr():
+            self.nvim.command(str(prev_winnr) + ' wincmd w')
 
     @neovim.autocmd("CursorMoved", pattern="*", eval="line('.')")
     def handleCursorMoved(self, line) -> None:
+        if self.nvim.current.buffer.options['buftype']:
```
autozimu: 👍  But please make a comment. It's not easy to understand what it does in first glance.

purpleP:
@autozimu You know what, I thought about it today and I think that maybe It's better to trigger User event instead when diagnostics change and let users do what they want with it, open quickfix, show notification or anything else. This would allow for example one interesting usecases that I actually use right now with custom linting plugin where I save errors in files for each linted file and this way I can show quickfix only when it's changed even between vim sessions.

PR #126, opened 2017-09-06
Open file and navigate to line:column in one step

hia3:
After `LanguageClient_textDocument_definition` you have to press Ctrl-O twice to return back where you started, this is not very convenient. This happens because it first opens the file, then moves the cursor.


autozimu:
Please also update test case.

autozimu:
Thanks.

PR #135, opened 2017-09-24
Add matched line content to ':Denite references'

cormacrelf:
```
Firstly by modifying the location dicts, and secondly by reading it off
again in the references plugin.

Then, adds the same highlighting as ':Denite grep' to the plugin. Also
copies the grep output format such that character ':0' is not rendered.
```

-----

It's not quite good to merge, I'm not very far into the codebase so I can't debug one thing, that only happens the first time you run it in a session. Repro:

1. Start LanguageClient-neovim. I'm using `sourcegraph/javascript-typescript-langserver`.
2. Find a token to find refs for and `Denite references` it.
3. Click through error output. Here is [my error output](https://pastebin.com/bYHhpM8u)
4. Do step 2 again, and it works from then on. 

I can't tell what's really happening behind that error (a keyboard interrupt? huh?). You can ignore that [Warning] readFile stuff, it happens on master branch too, it's just js-ts-langserver being silly.

Any ideas?

Edit: should refer to #81 since that's what this is implementing

**EDIT 2: fixed! That was just because I had :set shell=powershell. Never do that, folks!**

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -784,7 +784,7 @@ class LanguageClient:
                 start = loc["range"]["start"]
                 line = start["line"] + 1
                 character = start["character"] + 1
-                text = get_file_line(uri_to_path(loc["uri"]), line)
+                text = loc["text"]
                 entry = "{}:{}:{}: {}".format(path, line, character, text)
                 source.append(entry)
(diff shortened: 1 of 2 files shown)


autozimu:
Could you check travis output?

cormacrelf:
Done

autozimu:
Thanks.

PR #142, opened 2017-10-05
Only add/remove signs on a unique change

rschifflin:
Currently to add/remove signs to the signcol (to highlight lines with errors, warnings, etc), we're walking through the diff of `oldSigns` vs `newSigns` and performing the appropriate action. 

However, the signcol doesn't care if I have multiple of a sign for a line (perhaps an error on col 3 AND an error on col 17 of the same line). If I just eliminate one of the two errors, the diff is still a deletion, which removes the sign from the signcol entirely.

This fix instead compares a unique `oldSigns` vs a unique `newSigns` to only generate a diff action that changes the flat signcol.

### DIFF
--- a/rplugin/python3/LanguageClient/util.py
+++ b/rplugin/python3/LanguageClient/util.py
@@ -119,7 +119,6 @@ def retry(span, count, condition):
         time.sleep(span)
         count -= 1
 
-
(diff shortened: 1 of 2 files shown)


autozimu:
👍

PR #143, opened 2017-10-06
Use distinct sign ids when placing col signs

rschifflin:
In Vim, placing or unplacing a sign requires binding an id for that sign. Currently we use the line # of a sign as the id of that sign. This leads us to define the same id multiple times if, for instance, a single line has both a warning and an error. `unplace`ing that sign id 
 (for example, by solving that line's warning but not the error) then incorrectly clears both signs.

This change adds a simple calculation for determining an id for a given Sign. Note that we start numbering at 75000- an [arbitrarily-high number](https://github.com/airblade/vim-gitgutter/blob/master/autoload/gitgutter/sign.vim#L1) to attempt to avoid trampling other plugins' sign ids. We also only `unplace` within the sign's given buffer.

### DIFF
--- a/rplugin/python3/LanguageClient/DiagnosticsDisplay.py
+++ b/rplugin/python3/LanguageClient/DiagnosticsDisplay.py
@@ -1,23 +1,23 @@
 DiagnosticsDisplay = {
-    1: {
+    "1": {
         "name": "Error",
         "texthl": "ALEError",
         "signText": "✖",
         "signTexthl": "ALEErrorSign",
     },
-    2: {
+    "2": {
         "name": "Warning",
         "texthl": "ALEWarning",
         "signText": "⚠",
         "signTexthl": "ALEWarningSign",
     },
-    3: {
+    "3": {
         "name": "Information",
         "texthl": "ALEInfo",
         "signText": "ℹ",
         "signTexthl": "ALEInfoSign",
     },
-    4: {
+    "4": {
         "name": "Hint",
         "texthl": "ALEInfo",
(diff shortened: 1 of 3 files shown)


rplugin/python3/LanguageClient/DiagnosticsDisplay.py
```diff
@@ -1,23 +1,23 @@
 DiagnosticsDisplay = {
-    1: {
+    "1": {
```
autozimu: Thanks for the work.

Can you explain why this change is necessary?

I am asking because this change would break backward compatibility. User can [customize this display](https://github.com/autozimu/LanguageClient-neovim/blob/master/doc/LanguageClient.txt#L70).

rschifflin: Ah! I made that change because the integration tests read in the state (including the diagnostics) as a [json payload](https://github.com/autozimu/LanguageClient-neovim/blob/master/rplugin/python3/LanguageClient/LanguageClient_test.py#L47), which only supports string keys, and merges it with the default state containing numbered keys. So the merged end-result DiagnosticsDisplay was 
`{ 1: ..., 2: ..., 3: ..., 4: ..., "1": ..., "2": ..., "3": ..., "4": ... }`
which seemed unintended and threw off the total diagnostic count I was using to calculate sign ids.

I'll simply make a list of the `name`s of a DiagnosticDisplay (excluding duplicates) and use that to calculate an offset instead.

autozimu:
Thanks!

PR #151, opened 2017-10-24
Add vim support.

autozimu:


roxma:
hmm...

`nvim_buf_add_highlight` does not exist in vim-hug-neovim-rpc

And it seems this is a neovim-only feature.

autozimu:
In that case, I can probably detect current feature and do not call the API for vim, which is somewhat expected. Even if I managed to get this work on vim, some of the features might never get to the state of support for neovim.

roxma:
> I can probably detect current feature and do not call the API for vim

I can add it into vim-hug-neovim-rpc, and it will simply ignore the error for now. I'll update the documentation for this.

This feature is hard to workaround. But it only affects the views, so ignore the error should be just fine.

roxma:
Unfortunately I'm still having issue with completion with LanguageServer-php-neovim

screenshot_20171026_102946

screenshot_20171026_103035

These messages are forwarded from stderr of the process. But I'm not sure where it's coming from.

autozimu:
I cannot reproduce it.

Can you provide min vimrc and each version numbers?

roxma:
I'm debugging.

The error message forwarded by yarp is not well formatted. I'lll post detailed information when I fixed the message format.

roxma:
Unfortunately, due to vim's issue on handling the python `None`, I workaround this issue by converted it to empty string.

https://github.com/vim/vim/issues/2246

No error anymore.

autozimu:
👍

PR #152, opened 2017-10-25
Add ncm completion handler. Fix goto definition. add LanguageClientStart

roxma:
Goto definition works if I add a dummy function  `nvim_buf_add_highlight` into vim-hug-neovi-rpc to ignore the error.

#151


autozimu:
Cool, thanks!

PR #163, opened 2017-11-11
Implement signatureHelp

SevInf:
Echoes current signature , highlighting the currently active paramters.
If called on CursorMoved autocmd, can replace echodoc.

Screenshot: 
image

### DIFF
--- a/rplugin/python3/LanguageClient/LanguageClient.py
+++ b/rplugin/python3/LanguageClient/LanguageClient.py
@@ -16,7 +16,7 @@ from .TextDocumentItem import TextDocumentItem
 from .logger import logger, logpath_server, setLoggingLevel
 from .state import (
     state, update_state, execute_command, echo, echomsg, echoerr,
-    echo_ellipsis, make_serializable, set_state, alive)
+    echo_ellipsis, echo_signature, make_serializable, set_state, alive)
 from .util import (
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks.

njunxuan:
Is there an example on how this function can be implemented? Thank you!

DeltaEvo:
You can do it with this piece of vimscript
```vim
let b:Plugin_LanguageClient_started = 0

augroup LanguageClient_signature
	autocmd!
	autocmd User LanguageClientStarted let b:Plugin_LanguageClient_started = 1
	autocmd User LanguageClientStopped let b:Plugin_LanguageClient_started = 0
	autocmd CursorMoved * if b:Plugin_LanguageClient_started | call LanguageClient_textDocument_signatureHelp() | endif
augroup END
```

njunxuan:
Thanks I was able to get the function working in some situations. Though a string of errors frequently occur. I am guessing this is not integrated/ready for show time yet?

DeltaEvo:
What error do you get  ?

njunxuan:
So 2 things I've noticed.

1. When the function works, I have to exit out of insert mode and move my cursor before the signature appears. I am using this together with Deoplete.

2. Sometimes, i get a whole screen of errors. I tested with print() in python and it got me this:
screen shot 2017-11-16 at 11 09 31 am

Sorry about the screenshot, I'm not familiar with how to copy out the error yet :/

PR #164, opened 2017-11-11
Use buffer options for formatting

jungomi:
The formatting should use the `tabstop` and `expandtab` settings from the current buffer, not the global settings.

For example with this config:

```vim
set tabstop=2
set expandtab

autocmd FileType rust setlocal tabstop=4
```

The formatting of a Rust file should be done with `tabSize: 4`.


autozimu:
Thanks.

PR #165, opened 2017-11-11
Remove final empty line after formatting

jungomi:
Vim inserts a newline at the end of the file if `fixendofline` is set (set by default). When the formatting result includes a `\n` at the end, an empty line will be inserted and an additional `\n` is added, which means that the file ends with `\n\n`.

To avoid that, the final empty line is removed from the buffer if `fixendofline` is set.


autozimu:
Thanks.

Issue #166, opened 2017-11-12
codeAction command split, LSP Protocol, Haskell IDE Engine


meck:
I'm trying to get codeAction to work with [Haskell ide engine](https://github.com/haskell/haskell-ide-engine) and noticed its splitting the commands on ":"  where as the protocol calls for ": " colon space.

HIE has the command applyrefact:applyOne so that breaks. I've tried changing the split and that works, however the codeAction itself is still not working, but I've not yet had time to look any more in depth at that yet.

meck:
Seems HIE works fine with codeAction when I do this:
``    def fzfSinkTextDocumentCodeAction(self, lines: str) -> None:
        command, title = lines[0].split(": ")
``
in LanguageClient.py, I could do a PR if you want it, I'm not sure how other Language Servers would work in this regard...

autozimu:
PR is welcome.

PR #167, opened 2017-11-12
Changed split for CodeAction to include a space

meck:
As the protocol specifies that commands should be followed by a colon and a space, if there is a colon the command itself it breaks as  the issue [here](https://github.com/autozimu/LanguageClient-neovim/issues/166) with the haskell language server.


autozimu:
Thanks.

PR #174, opened 2017-11-15
add try except in rpc header handling

stCarolas:
Because wrong headers  break "while not self.infile.closed" cycle silently
Example: [vscode-javac](https://github.com/georgewfraser/vscode-javac) sends `pattern: glob:/home/stcarolas/.gradle/caches/modules-*/files-*/weblogic/javax.ejb/3.0.1/*/javax.ejb-3.0.1.jar`

### DIFF
--- a/rplugin/python3/LanguageClient/RPC.py
+++ b/rplugin/python3/LanguageClient/RPC.py
@@ -62,9 +62,13 @@ class RPC:
         while not self.infile.closed:
             line = self.infile.readline().decode("UTF-8").strip()
             if line:
-                header, value = line.split(":")
-                if header == "Content-Length":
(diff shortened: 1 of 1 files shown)


autozimu:
I don't get it.

if the rogue line is `pattern: glob:/home/stcarolas/.gradle/caches/modules-*/files-*/weblogic/javax.ejb/3.0.1/*/javax.ejb-3.0.1.jar`, after split, the `header` will be `pattern`, which does not equal to `Content-Length`, thus the program will continue to next loop. I don't see why exception could be thrown in this case.

Please correct me if I'm wrong.

autozimu:
Closing as no response.

Issue #178, opened 2017-11-18
How to set InitializeParams.initializationOptions?


htfy96:
I am trying to get this client working with [CQuery](https://github.com/jacobdufault/cquery), a new C++ language server. However, it reported that `FATL| Initialization parameters (particularily cacheDirectory) are required` in its log because during `initialize`, `initializationOptions.cacheDirectory` must be set. 

It seems that currently there is no way to set these options in this plugin. This problem also affects `clangd`, which by default works only in C++98 mode and considers all syntax in C++11 as errors. Is it possible to set `initializationOptions` with default values/callback functions?

Thanks.

autozimu:
Resolved in ef2b82c41a864e8b26e58897e4bb7f4dd5576cd2.

DeltaEvo:
I added a little wiki page to help to settup cquery with LanguageClient-neovim https://github.com/autozimu/LanguageClient-neovim/wiki/cquery

akinh:
Was trying to fix this issue, and updated my autozimu/LanguageClient-neovim repo. However, getting 

```
error caught in async handler '/home/username/.vim/bundle/LanguageClient-neovim/rplugin/python3/LanguageClient:autocmd:CursorMoved:* [['', 56]]'
Traceback (most recent call last):
  File "/home/username/.vim/bundle/LanguageClient-neovim/rplugin/python3/LanguageClient/LanguageClient.py", line 1171, in handle_CursorMoved
    buftype, line = gather_args(["buftype", "line"], args=args)
  File "/home/username/.vim/bundle/LanguageClient-neovim/rplugin/python3/LanguageClient/LanguageClient.py", line 75, in gather_args
    if len(args) > 1 and len(args[1]) > 0:  # from vimscript side
TypeError: object of type 'int' has no len()
```
this error now. Any ideas how to fix ?

akinh:
UpdateRemotePlugins fixed it. Sorry for the noise.

akinh:
Still getting the same error after updating the plugin and setting the params in .vim/settings.json. When I run cquery from command line, it works fine though.

MaskRay:
It's better to have some way to use different initializationOptions for different language servers

```
{
        "initializationOptions": {
                "cacheDirectory": "/tmp/cquery",
		"resourceDirectory": "yourCQueryDirectory/clang_resource_dir"
        }
}
```

oblitum:
[...]

PR #179, opened 2017-11-18
Add settings file support

### LINKED ISSUE #139: Settings file support
Am I correct in assuming that settings files like this: https://github.com/rust-lang-nursery/rls#configuration are not supported in LanguageClient-neovim? If so could you give me a few pointers as to what it would take to implement this.

DeltaEvo:
closes #139, #178


autozimu:
Thanks.

PR #185, opened 2017-11-24
Also look for ".vim" when searching for the project root directory.

xbreak:
This improves the situation for developers using vim and e.g. C++ where there is no clear root project file, like "Cargo.toml" for rust, and especially the case when they use Subversion because the checkout and thus the location of ".svn" is often at a higher level, encompassing multiple projects.

### DIFF
--- a/rplugin/python3/LanguageClient/util.py
+++ b/rplugin/python3/LanguageClient/util.py
@@ -58,6 +58,7 @@ def get_rootPath(filepath: str, languageId: str) -> str:
         rootPath = traverse_up(
             filepath,
             lambda folder: (
+                os.path.exists(os.path.join(folder, ".vim")) or
                 os.path.exists(os.path.join(folder, ".git")) or
                 os.path.exists(os.path.join(folder, ".hg")) or
                 os.path.exists(os.path.join(folder, ".svn"))))


DeltaEvo:
I don't wanted to add it when I added settings file support because:
 - It could conflict with user's `.vim` used to store vim config
 - We don't use editor/ide settings folder as root marker

What do you think about this @autozimu ?

@xbreak in your c++ projet do you have a Makefile or a CMakeList.txt or any build configuration file ? Because this could be a root marker that I think could be added into the `get_rootPath` function

xbreak:
@DeltaEvo: 
We don't use an editor/settings file as a root marker either, which is the flawed assumption by the plugin currently I guess, only to fall back to another assumption on SCM working copy detection.  
With the path of least resistance I adopted the plugin default `.vim/settings.json` path to configure the language server (cquery) from the project root which makes sense because each project could be configured individually. 

But really this should then be fully user configurable since there are too many ways of structuring a project. Maybe a a configuration variable that consists of patterns to search upwards for?

autozimu:
Like DeltaEvo mentioned, do your projects has a feature to be detected as project root?

At a last resort, you can always specify project root using something like `:LanguageClientStart rootPath=/path/to/dir`.

Letting user specify rules to identify project root, IMO, just sounds great on paper, but simply won't work in reality.

DeltaEvo:
@autozimu Why it will won't work in reality ?

A lot of plugins that have automatic root detection have a setting for that
https://github.com/dbakker/vim-projectroot/blob/master/doc/projectroot.txt#L192-L198
https://github.com/airblade/vim-rooter/blob/master/doc/rooter.txt#L83-L89
https://github.com/kien/ctrlp.vim/blob/master/doc/ctrlp.txt#L198-L203

xbreak:
@autozimu: Not really. We use waf, but from the system (no project level waf script) so our projects simply contain a single `wscript` file at the root and in each subdirectory.
So I would have to provide my own marker, which is not a problem in this case because I have to provide the `settings.json` for cquery somewhere in the root anyway.

autozimu:
> it will won't work in reality

I was thinking the approach to let user specify a predicate function to mark project root won't work.

The approach of using customizable markers does seem feasible.

xbreak:
I'll close this PR then. Should I create a feature request for a "`g:LanguageClient_rootmarkers`" variable (and maybe a "`g:LanguageClient_projectroot`" that is default None but can be used to force a path)?

autozimu:
Please make it `g:LanguageClient_rootMarkers`, thanks.

PR #188, opened 2017-11-30
Don't assume rpc message is always valid UTF8

yshui:
Ignore invalid message instead of crashing

### DIFF
--- a/rplugin/python3/LanguageClient/RPC.py
+++ b/rplugin/python3/LanguageClient/RPC.py
@@ -60,13 +60,25 @@ class RPC:
     def serve(self):
         content_length = 0
         while not self.infile.closed:
-            line = self.infile.readline().decode("UTF-8").strip()
+            try:
+                data = self.infile.readline()
+                line = data.decode("UTF-8").strip()
+            except UnicodeError:
+                msg = "Failed to decode message as UTF-8: " + str(data)
+                logger.exception(msg)
+                continue
             if line:
                 header, value = line.split(":")
                 if header == "Content-Length":
                     content_length = int(value)
             else:
-                content = self.infile.read(content_length).decode("UTF-8")
+                try:
+                    data = self.infile.read(content_length)
+                    content = data.decode("UTF-8")
+                except UnicodeError:
+                    msg = "Failed to decode message as UTF-8: " + str(data)
+                    logger.exception(msg)
+                    continue
                 logger.debug("<= " + content)
                 try:
                     msg = json.loads(content)


rplugin/python3/LanguageClient/RPC.py
```diff
@@ -60,13 +60,25 @@ def notify(self, method: str, params: Dict[str, Any]) -> None:
...
                 header, value = line.split(":")
                 if header == "Content-Length":
                     content_length = int(value)
             else:
-                content = self.infile.read(content_length).decode("UTF-8")
+                try:
+                    data = self.infile.read(content_length)
+                    content = data.decode("UTF-8")
+                except Exception:
+                    msg = "Failed to decode message as UTF-8: " + data
```
autozimu: `+` cannot be used to concatenate string and bytes.

rplugin/python3/LanguageClient/RPC.py
```diff
@@ -60,13 +60,25 @@ def notify(self, method: str, params: Dict[str, Any]) -> None:
...
             if line:
                 header, value = line.split(":")
                 if header == "Content-Length":
                     content_length = int(value)
             else:
-                content = self.infile.read(content_length).decode("UTF-8")
+                try:
+                    data = self.infile.read(content_length)
+                    content = data.decode("UTF-8")
+                except Exception:
```
autozimu: Instead of catching all exception here, would it be better catching `UnicodeError`?

rplugin/python3/LanguageClient/RPC.py
yshui: So what should be done here?

autozimu: Could use `str(data)`

yshui: I changed it to data.decode('utf-8', 'backslashreplace')

autozimu: I don't think change to `data.decode('utf-8', 'backslashreplace')` will bring any value here.

The line is in a catch clause. What if `data.decode('utf-8'), 'backslashreplace'` still decode error and throws an exception? On the other hand, if the `backslashreplace` decode successfully, the benefit is the log message is more readable, but the whole read content operation already failed, and that readable log message will further obscure real issue.

yshui: Done

yshui:
If the exception handler is call, the main thread seems to go into a dead loop.

Not sure why

yshui:
OK, the main thread just goes into a dead loop on its own sometimes. So I think it's unrelated.

autozimu:
Yes, it's a different issue.

FYI, I will deprecate this python implementation soon with https://github.com/autozimu/LanguageClient-neovim/tree/next.

autozimu:
Thanks.

PR #191, opened 2017-12-05
Don't overload get_complete_position

yshui:
The default one is just fine. Adding a `*` risks triggering exponential behaviour in Python re.

I don't quite see the point of adding a `*`? But if that's really necessary, we should move to a different regex implementation.


autozimu:
Thanks.

PR #195, opened 2017-12-06
Signature help minor fixes

phil-nelson:
I noticed a couple of bugs when using signature help:

* If there's no active parameter it can crash because of a missing return
* In PHP function parameters can have a `'` character in them which need to be escaped to display properly

Off topic but it would be really nice to have a popup which displayed signature help, triggered by the triggerCharacters, similar to what vscode does. I had a quick look and it doesn't look straightforward to do this with neovim though.


autozimu:
Thanks.

Yes, it is not easy to achieve IDE like parameter help with stock neovim/vim.

PR #201, opened 2017-12-10
Specify the next branch explicitly.

piouPiouM:
According https://github.com/junegunn/vim-plug/issues/288#issuecomment-151896611 we have to precise the branch to clone if different of "master" in despite the fact that the "next" branch is set by default:

> `Plug 'something'` is simply a shorthand notation of `Plug 'something', { 'branch': 'master' }`

I was trying many times to install LanguageClient-neovim from scratch and invariably the master branch was cloned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,10 @@ functional.)
 Using [`vim-plug`](https://github.com/junegunn/vim-plug):
 
 ```vim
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks.

PR #205, opened 2017-12-11
use `-f` instead of `--force`

euclio:
This is compatible with both GNU and BSD `cp`, which is what Mac OS uses.


autozimu:
Thanks.

PR #237, opened 2017-12-27
Prevent making vim unusable in case we are loaded in an env that does not have yarp#py3 defined.

akatrevorjay:
There are unfortunately a multitude of different methods that this can be the case, but really any time you run without a working python env (eg yet/anymore), you'll find yourself an unusable vim/nvim because `yarp#py3` doesn't exist, and hooks are loaded to `HandleCursorMoved` and such ;)

This just checks for it before setting up hooks on everything.

### DIFF
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -5,7 +5,7 @@ endif
 if !exists('yarp#py3')
     echoerr 'LanguageClient: yarp#py3 does not exist. Refusing to load.'
     finish
-fi
+endif
 
 command LanguageClientStart call LanguageClient_start()
 command LanguageClientStop call LanguageClient_stop()
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -2,6 +2,11 @@ if has('nvim')
     finish
 endif
 
+if !exists('yarp#py3')
+    echoerr 'LanguageClient: yarp#py3 does not exist. Refusing to load.'
+    finish
+fi
+
 command LanguageClientStart call LanguageClient_start()
 command LanguageClientStop call LanguageClient_stop()
 
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -2,7 +2,7 @@ if has('nvim')
     finish
 endif
 
-if !exists('yarp#py3')
+if !exists('*yarp#py3')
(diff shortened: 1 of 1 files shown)


akatrevorjay:
Wait, what? How did these other commits get placed here?!?

akatrevorjay:
Uh, non mergable? is appveyor confused since I switched branches in post?

plugin/LanguageClient.vim
```diff
@@ -2,6 +2,11 @@ if has('nvim')
     finish
 endif
 
+if !exists('yarp#py3')
+    echoerr 'LanguageClient: yarp#py3 does not exist. Refusing to load.'
+    finish
+fi
```
autozimu: Should be `endif` instead of `fi`.

akatrevorjay: Lol. Whoops, guess I was in bash mode.

akatrevorjay: Done btw

plugin/LanguageClient.vim
```diff
@@ -2,6 +2,11 @@ if has('nvim')
     finish
 endif
 
+if !exists('yarp#py3')
```
autozimu: Should be `*yarf#py3` instead of `yarf#py3`.

akatrevorjay: Sorry, can you explain that one further? I think the formatting is messed in email.

autozimu: Sorry for the typo. I meant yarp.

For chenking function existence in vimscript, you will need to use *funcname, instead of just funcname. Check doc by :h exists

If the formatting is messsed up in email, check it on github

akatrevorjay: Interesting. I wonder how I never ran into that before. Fixed!

matthiasbeyer:
Can someone tell me what this `yarp` thing is? I am trying this plugin on NixOS and am running into this issue because the package for LanguageClient-neovim seems not to pull in this dependency...

autozimu:
It's a dependent plugin of the original python implementation (only needed for vim users). https://github.com/autozimu/LanguageClient-neovim/blob/master/INSTALL.md#30-vim-only

I'd suggest you switch to rust implementation (branch next), which don't have any dependency.

matthiasbeyer:
Thanks. Filed an issue in nixpkgs for this: https://github.com/NixOS/nixpkgs/issues/33391

PR #243, opened 2017-12-28
Expose RLS and cquery build progress

### LINKED ISSUE #235: Show build progress
Language servers like RLS and cquery provide custom notifications for the build progress which could be useful to display (to show that the language server isn't showing the diagnostics because it's doing something as opposed to being broken or whatever, and to show the indexing/build progress if the language server like cquery provides it).

I implemented simple `echo` for the three RLS notifications in [this branch](https://github.com/YaLTeR/LanguageClient-neovim/tree/rls-progress). From the initial testing, they are nice on the first file open but get spammy as you type. I think showing them only the first time would work, however some kind of a dedicated indicator would be better. I'm not sure how to make one though as I'm not experienced with vim plugin writing.

Making an issue to ask for feedback and ideas on how to show this.

YaLTeR:
Resolves #235.

This PR adds two global variables:
- `g:LanguageClient_building` is 1 when the language server is building something and 0 otherwise,
- `g:LanguageClient_buildStatus` is a short string describing the build status (`Rust: build started`, `cquery: indexing (5 jobs)`, `cquery: idle`, etc).

It also adds a function, `LanguageClient_statusLine()`, which returns either an empty string or a message like `[Rust: build completed]` which can be put into the statusline.

The variables are updated by listening to RLS's and cqeury's build progress notifications. It's easy to extend the functionality to other language servers.

There's one issue with the implementation: cquery's build progress messages arrive very quickly and due to each message being processed in a separate thread, the next-to-last progress message frequently ends up being processed last and the build status remains set to "building". Not sure what the best way of addressing that would be.

### DIFF
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -581,13 +581,13 @@ function! LanguageClient_exit() abort
                 \ })
 endfunction
 
-" Set to 1 when the language server is building something.
-let g:LanguageClient_building = 0
+" Set to 1 when the language server is busy (e.g. building the code).
+let g:LanguageClient_serverBusy = 0
 function! LanguageClient_statusLine() abort
-    if !exists('g:LanguageClient_buildStatus')
+    if !exists('g:LanguageClient_serverStatusMessage')
         return ''
     endif
-    return '['.g:LanguageClient_buildStatus.']'
+    return '['.g:LanguageClient_serverStatusMessage.']'
 endfunction
 
 " When editing a [No Name] file, neovim reports filename as "", while vim reports null.
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -581,6 +581,15 @@ function! LanguageClient_exit() abort
                 \ })
 endfunction
 
+" Set to 1 when the language server is building something.
+let g:LanguageClient_building = 0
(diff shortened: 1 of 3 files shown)


autozimu:
👍  Good work.

Only minor issue. The variable names contain `build` while not every language has a definition of `build`. IMO, names `g:LanguageClient_serverStatus` and `g:LanguageClient_serverStatusMessage` are more appropriate here. Also, you can name the corresponding functions same as variables to reduce confusion, with previous mentioned proposal, they would be `LanguageClient_serverStatus()` and `LanguageClient_serverStatusMessage()`.

YaLTeR:
Rebased and renamed (building -> serverBusy, buildStatus -> serverStatusMessage).

autozimu:
Thanks. I will merge this in but will make small modification shortly.

YaLTeR:
Is it actually needed to handle each message in a separate thread? Maybe just have one handler thread running, and send messages to it from the reading thread via channels? That will solve the cquery issue I outlined and potentially others since the messages from the language server will be processed in order.

autozimu:
It is definitely possible.

But in current state, there is no easy and reliable to do async work in rust, especially with stable rust.

If using one thread to handle requests sequentially, one time-consuming request will block all the following ones, which is probably not favorable by end user.

PR #245, opened 2017-12-28
Fix warning about unknown project type

mkollaro:
Result.or() gets executed even when the value is Result::Ok, which
causes the warning about "Unknown project type" to show even though the
project directory was successfully found. It also unnecessarily
traverses the directory tree in search for VCS files.


autozimu:
Thanks.

PR #250, opened 2017-12-31
Provide a sync version of textDocument_hover

yshui:
In neovim, async functions provided by remote plugins can't return a value, render the result unusable.


autozimu:
Could you check CI error? Seems like linting issue.

PR #264, opened 2018-01-06
fix duplicate LanguageClient_notify helptag

slogsdon:


autozimu:
Could you update to LanguageClient#Call ?

slogsdon:
Will do. Was originally trying to match existing helptags

slogsdon:
@autozimu Updated

autozimu:
thanks

PR #280, opened 2018-01-15
Add missing CompletionItemKind

yshui:
Reference: https://microsoft.github.io/language-server-protocol/specification


autozimu:
Thanks.

PR #285, opened 2018-01-17
Add execute bits to build binary.

yuzu441:


autozimu:
Can you explain the reason a bit?

I believe the binary compiled by cargo is already having executable bits. Maybe I missed something here?

yuzu441:
Sorry, I sent PR by mistake, but my build binary by cargo don't have execute bits.

PR #286, opened 2018-01-17
Fix replacing placeholders in completion text in deoplete plugin

### LINKED ISSUE #232: Function parameter expansion with nvim-completion-manager and UltiSnips?
- neovim version: 0.2.2
- Plugin version (git SHA): 640a74cce174172245429b3096bf3bd88053dbcf
- Language server name/link and version: cquery or rls

Is there a way of getting the function parameters in completed items to behave like they would in snippets? Right now upon completing a function I get a huge line like:
```cpp
printf(${1:const char *__restrict __format, ...})$0
```
and have to clean it up by myself. If I understand correctly, in this case NCM should use UltiSnips to placeholder the function parameters with their names and allow me to jump between them with the UltiSnips bindings.

Maybe there's some setup I'm missing?

nowrep:
simplify_snippet is no longer called since 5228816283

I'm not sure how correct this code is, but it works for me. May also fix #232

### DIFF
--- a/rplugin/python3/deoplete/sources/LanguageClientSource.py
+++ b/rplugin/python3/deoplete/sources/LanguageClientSource.py
@@ -48,4 +48,7 @@ class Source(Base):
         result = self.vim.funcs.eval("remove({}, 0)".format(CompleteResults))
         if not isinstance(result, list):
             result = []
+        else:
+            for item in result:
+                item["word"] = simplify_snippet(item["word"])
         return result


autozimu:
Thanks for contributing.

The original solution is a hacky way to solve the problem.

The way it should work is that server should not send snippet when client declares that it doesn't support, like this commit in rls https://github.com/rust-lang-nursery/rls/pull/648. If your language server of choice doesn't support this, file an issue against them.

And https://github.com/autozimu/LanguageClient-neovim/issues/232 is more about using snippet when it is supported, so having this merged in won't help it.

nowrep:
Well, this actually does fix snippets with deoplete + neosnippet.

You are right, it won't help #232 as that is reported when using with ncm and not deoplete.

nowrep:
> The way it should work is that server should not send snippet when client declares that it doesn't support, like this commit in rls rust-lang-nursery/rls#648. If your language server of choice doesn't support this, file an issue against them.

LanguageClient-neovim does request snippets if it detects snippet plugin installed (`s:hasSnippetSupport()`) but then it fails to convert the syntax so the snippet plugin understands it. So the bug in this case is in LanguageClient-neovim and not the language server.

autozimu:
Yes, and it’s different topic than this PR.

PR #300, opened 2018-01-28
Add includeDeclaration to textDocument/references

MaskRay:
If users want to include declarations:

```vim
nn <M-,> :call LanguageClient_textDocument_references({'includeDeclaration': v:true})<cr>
```


autozimu:
Thanks.

Issue #281, opened 2018-01-16
Deoplete integration bug with complexes projects on TypeScript


JulioJu:
- neovim/vim version NVIM v0.2.3-450-g60d6a8b132.3-450-g60d6a8b13
- Plugin version (git SHA)  b4033ed 
- Minimal vimrc
```vim
call plug#begin('~/.vim/plugged')
" Syntax typescript
Plug 'leafgarland/typescript-vim', { 'for': ['typescript']}
Plug 'HerringtonDarkholme/yats.vim', { 'for': ['typescript']}
Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }
Plug 'Shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }
call plug#end()
filetype plugin indent on
let g:deoplete#enable_at_startup = 0
autocmd InsertEnter * call deoplete#enable()
let g:LanguageClient_serverCommands = {
    \ 'typescript': ['javascript-typescript-stdio'],
    \ 'javascript': ['javascript-typescript-stdio'],
    \ }
let g:LanguageClient_loggingLevel = 'DEBUG'
set nu
```
- Language server name/link and version `javascript-typescript-stdio --version`: 2.5.5

- Reproduction steps (from clean state)
    1. Download [smallProjectTSForBugReportOnLanguageClient.zip](https://github.com/autozimu/LanguageClient-neovim/files/1634157/smallProjectTSForBugReportOnLanguageClient.zip)
    2. Unzip previous zip file
    3. `npm install`
    4.  `nvim -u ~/minimalVimrc app/server.js app/server.ts`
    5. On line 15, type "console." and wait. You could see only a popup menu with « log [M] ». This menu is gathering only by Deoplete.
    6. If you type « l » and wait a long time  on a slow computer ( Intel Core i5 M 520 (2.4 GHz, 2010), 8Go Ram) you could see « log f [LC] » gathering by LanguageClient. 
    7. Now if you type one time on the backspace key, you could see the LanguageClient completion menu appears. 
    8. On more complex projects, we could see than the LanguageClient completion menu and Deoplete completion menu never cooperates. I've seen sometime one appears before the other, or the menu blink… 
    
- Logs
[LanguageServer.log](https://github.com/autozimu/LanguageClient-neovim/files/1634219/LanguageServer.log)
[...]

PR #310, opened 2018-02-09
Clear cache before creating completion candidates in deoplete.

yymm:
I encountered a problem in which completion candidates do not work out if I type Backspace during completion similar to #281.

Since it seemed that the previous completion candidates remained, I cleared the previous completion result each time we create an completion candidate.

Thank you.


autozimu:
Thanks for contributing.

The problem is that the variable is shared. Resetting it would break other/futher extension.

We probably need to properly cleanup the state in case of deoplete error, or make the variable unique.

yymm:
I see, it's better to deal with deoplete's source.

I will PR again when I get a good idea :)

PR #318, opened 2018-02-27
fix powershell installation script

yatli:


autozimu:
What issue was fixed with this change?

yatli:
hi @autozimu!

`install.ps1` wasn't working correctly -- the url is https but the default policy uses wrong Tls protocol. Both Invoke-WebRequest and the WebClient would throw an exception.

autozimu:
Hmm, just curious, which version of Windows is it when the script throws
exception?

I tested the original script in Windows7 VM and didn’t see any exception.

yatli:
`Windows10 Version 1709 build 16299`.
The arch is `x64`.

autozimu:
I see and thanks for the prompt response.

I will merge this in after verifying it still works in Windows7.

yatli:
if Invoke-WebRequest wasn't in win7 please use the WebClient solution.
btw check this out: https://www.ssllabs.com/ssltest/analyze.html?d=github.com&s=192.30.255.112&hideResults=on

Perhaps Github had not switched to Tls1.2 when you tested it out.

yatli:
update: WebClient does not work with the ServicePointManager fix.

autozimu:
image

Please fix for Windows 7.

yatli:
@autozimu got it. will fire up a vm

yatli:
but the problem is: now that Github only supports Tls12, what are we going to do with Win7? Have you tested it with your original script?

autozimu:
Hmm, good question.

Actually the original script also failed in Windows 7.

yatli:
oh well.. https://archive.codeplex.com/?p=win2k8r2tls12

autozimu:
It looks [Net 3.5 doesn't support TLS 1.2](https://msdn.microsoft.com/en-us/library/system.net.securityprotocoltype(v=vs.90).aspx) and it's been been end of support for quite some time. Users using Net 3.5 should upgrade Net framework either way.

Will merge this in shortly. Thanks.

Palpatineli:
I'm sorry if it sounds dumb but does it solve the windows 7 installation problem? I'm seeing the exact same problem shown in your screenshot. I'm using the current master as of 8/4/2018 and I'm on .net 4. windows 7 64bit sp1.

PR #321, opened 2018-03-04
Apply overlapping TextEdits in reverse order

idanarye:
LanguageClient-neovim adheres the LSP rules, but [some clients don't](https://github.com/eclipse/eclipse.jdt.ls/issues/571) - saying that [overlapping edits are OK in some "de facto standard"](https://github.com/eclipse/eclipse.jdt.ls/issues/571#issuecomment-370225470). To be compatible with these servers, some clients match the coincidental behavior of the VSCode implementation - applying overlapping edits in reverse.

[The old Python implementation of LanguageClient-neovim did it](https://github.com/autozimu/LanguageClient-neovim/blob/f47dd62/rplugin/python3/LanguageClient/LanguageClient.py#L168-L171), and the Rust implementation should do it too.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -195,9 +195,16 @@ pub trait ILanguageClient: IVim {
     fn apply_TextEdits<P: AsRef<Path>>(&self, path: P, edits: &[TextEdit]) -> Result<()> {
         debug!("Begin apply TextEdits: {:?}", edits);
         let mut edits = edits.to_vec();
-        edits.reverse();
+
+        // Edits must be applied from bottom to top, so that earlier edits will not interfere
+        // with the positioning of later edits. Edits that start with the same position must be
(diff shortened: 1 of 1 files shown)


idanarye:
Don't merge this yet - apparently [the current behavior is according to the specs](https://github.com/eclipse/eclipse.jdt.ls/issues/571#issuecomment-370234670), and all these other clients (including VSCode!) aren't?

idanarye:
The specs were fixed, so that the correct behavior is what VSCode and the other plugins use:

* https://github.com/Microsoft/vscode/commit/bde23fb906c97cec1949ca705b299a47a73b6f11#r27981287
* https://github.com/Microsoft/language-server-protocol/commit/b70345d

I've amended the comments in the PR, and it can now be merged.

autozimu:
Thanks so much for your effort 👍

PR #332, opened 2018-03-14
Fix file path calculation when using FZF

cynix:
When a list of locations are received from language server, relative
path is calculated from getcwd() before passing to FZF for selection.
However, when a selection is returned, the relative path is appended
to project root rather than cwd, which will result in the wrong file
being opened. Fix this by appending relative path to cwd instead.


autozimu:
Thanks.

PR #333, opened 2018-03-18
Fix the `g:LanguageClient_diagnosticsList` setting:

idanarye:
- Allow setting it to `v:null` for no-value
- Fix the documentation to match the `next` implementation:
  - `"Quickfix"` instead of `"quickfix"`
  - `"Location"` instead of `"locations"`
  - `v:null` instead of `""`

Since `next` is already using different values from `master`, I thought it was OK to use `v:null` instead of an empty string. Vim has `v:null` since Vim 8 - and earlier versions can't run LanguageClient-neovim anyways - and Neovim has it [since January 2016](https://github.com/neovim/neovim/commit/d70a322c40e849f98ad573d2a37dc680c5616b26) - and I doubt any Neovim user still uses a version that old.

Support the same values `master` supports requires some more work, but it's doable. If that's what you want say the word and I'll change this PR to do it.


autozimu:
Thanks.

Issue #329, opened 2018-03-12
Deoplete poorly support Rust completion via LC (trigger character)


andreyorst:
Hello. I've read from your [README.md](https://github.com/autozimu/LanguageClient-neovim/blob/next/README.md) that you support deoplete, and nvim-completion-manager (wich i've never heared of, but we'll come to that later).
Personally I use deoplete for a long time with [deoplete-clang](https://github.com/zchee/deoplete-clang) and it provides great completion system to me when I program in C and C++. I've decided to try Rust, and went through many different plugins that _provide_ completion system for it. But none of them use deoplete _(except deoplete-rust, but I couldn't make it to work. And the last commit is 8 month ago, so i think it's dead anyway)_, wich I use daily and don't want to change to the other thing for another language. If I could. And assuming that you support deoplete I think that I could. But there is an issue:

That's how deoplete works with C++:
![deoplete_clang](https://user-images.githubusercontent.com/19470159/37322714-57d2907c-2690-11e8-8293-1c8e84bede87.gif)

And this is how deoplete works with LanguageClient-neovim:
![deoplete_lc](https://user-images.githubusercontent.com/19470159/37322718-615e022a-2690-11e8-9370-7a2796708d7a.gif)

I type use std:: and nothing happens. Yes it will eventually show something if I type in some more, but I don't know Rust, and I like to check what completion system shows, to explore method names etc.
![image](https://user-images.githubusercontent.com/19470159/37296079-ecb3e6d4-262a-11e8-9346-b2875fab92f9.png)

Well but this is not what I guessed when watched gifs in readme... I started to think that's the plugin itself works this way, but I decided to give a try to nvim-completion-manager, and the result is fascinating:
![ncm_lc](https://user-images.githubusercontent.com/19470159/37322725-6a7e3866-2690-11e8-9244-9e9f80f74a53.gif)

I tried to use nvim-completion-manager for C/C++, but I don't quite like its completion output. When I switched from YCM to deoplete I feeled major difference, that deoplete provides better autocompletion list.

So I tried to reproduce it with theese plugins only: `deoplete + LanguageClient-neovim`. Still poor completion
However `nvim-completion-manager + LanguageClient-neovim` works wery well. Is there something wrong in my particular deoplete setup, or is it partly supported?

---

**neovim version:**
NVIM v0.2.2
Build type: Release
**Plugin version (git SHA):**
git rev-parse origin/master
[...]

PR #339, opened 2018-03-22
Fix deoplete completion issue.

yymm:
Fix Backspace problem(#281).
It solved by separating deoplete completion function(LanguageClient_omniComplete) and vim's omni complete function(LanguageClient#complete) in plugin/LanguageClient.vim, and completion results are cleared for each completion in deoplete source.

And fix input_pattern problem(#329), but problems still remain.
The current status can be complemented with words and complemented by "." And "::" and "->", but it does not correspond to other patterns.

There is no problem with Rust and Vue, JavaScript and TypeScript which I am using completion, but it may not work in other languages.I think it is necessary to respond separately for each language.

Since I can not solve this problem alone, I will set up an Issue or Pull Requrst separately.

### DIFF
--- a/plugin/LanguageClient.vim
+++ b/plugin/LanguageClient.vim
@@ -551,7 +551,7 @@ function! LanguageClient_NCMRefresh(info, context) abort
     return LanguageClient#Notify('LanguageClient_NCMRefresh', [a:info, a:context])
 endfunction
 
-let g:LanguageClient_completeResults = []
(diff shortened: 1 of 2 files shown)


autozimu:
Looks good. Thanks.

This should cover most common programming language trigger characters. Missing language support can be added later case by case.

yymm:
Thank you for review and marge.

> This should cover most common programming language trigger characters. Missing language support can be added later case by case.

OK.

PR #350, opened 2018-03-28
Fix newlines being trimmed from a MarkedString

YaLTeR:
`.collect()` was collecting into a `String`, not a `Vec`.


autozimu:
Thanks.

PR #351, opened 2018-03-28
Make sure bin/languageclient is always executable if built locally

ahmedelgabri:
`make release` was missing making `bin/languageclient` executable which shows this error when you open neovim

```sh
$ nvim
LanguageClient: Not executable!
Press ENTER or type command to continue
```


autozimu:
Thanks.

PR #353, opened 2018-03-29
Handle new RLS build notifications

YaLTeR:
With Rust 1.25.0, RLS delivers new (more interesting) progress messages.

The notification name seems general enough though, so instead of outputting `Rust: <something>` I decided to output `LS: <something>` for "LanguageServer".

I tested this and it seems to work well.

I'm yet to see it output `params.percentage` though, it's specified in their [docs](https://github.com/rust-lang-nursery/rls/blob/master/contributing.md#extensions-to-the-language-server-protocol) and there's a reference in the [source](https://github.com/rust-lang-nursery/rls/blob/f5a0c91a39368395b1c1ad322e04be7b6074bc65/src/build/plan.rs#L499-L502) but I'm not sure when that happens.

The comma in the `MarkedString` was removed by `cargo fmt`, I can strip that change away if needed.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -2074,45 +2074,40 @@ pub trait ILanguageClient: IVim {
         Ok(())
     }
 
-    fn window_handleProgress(&self, params: &Option<Params>) -> Result<()> {
+    fn window_progress(&self, params: &Option<Params>) -> Result<()> {
         info!("Begin {}", NOTIFICATION__WindowProgress);
         let params: WindowProgressParams = serde_json::from_value(params.clone().to_value())?;
 
-        let busy = !params.done.unwrap_or(false);
+        let done = params.done.unwrap_or(false);
 
         let mut buf = "LS: ".to_owned();
 
-        if busy {
+        if done {
+            buf += "Idle";
+        } else {
             // For RLS this can be "Build" or "Diagnostics" or "Indexing".
-            buf.push_str(
-                params
-                    .title
-                    .as_ref()
-                    .map(|title| title.as_ref())
-                    .unwrap_or("Busy"),
-            );
+            buf += params
+                .title
+                .as_ref()
+                .map(|title| title.as_ref())
+                .unwrap_or("Busy");
 
             // For RLS this is the crate name, present only if the progress isn't known.
             if let Some(message) = params.message {
-                buf.push_str(&format!(" ({})", &message));
+                buf += &format!(" ({})", &message);
             }
             // For RLS this is the progress percentage, present only if the it's known.
             if let Some(percentage) = params.percentage {
-                buf.push_str(&format!(" ({:.1}% done)", percentage));
+                buf += &format!(" ({:.1}% done)", percentage);
             }
-        } else {
-            buf.push_str("Idle");
         }
-
-        // Sanitize before sending to vim.
-        buf.replace('\'', "''");
 
         self.command(&format!(
             "let {}={} | let {}='{}'",
             VIM__ServerStatus,
-            if busy { 1 } else { 0 },
+            if done { 0 } else { 1 },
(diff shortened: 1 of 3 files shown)


src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
         Ok(())
     }
 
+    fn window_handleProgress(&self, params: &Option<Params>) -> Result<()> {
```
autozimu: Function name should be `window_progress`.

src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
...
+        info!("Begin {}", NOTIFICATION__WindowProgress);
+        let params: WindowProgressParams = serde_json::from_value(params.clone().to_value())?;
```
autozimu: `.to_value()` is redundant.

src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
...
+
+        let busy = !params.done.unwrap_or(false);
```
autozimu: The logic is convoluted here. Simply use `let done = ...` can make the logic clearer.

src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
...
+
+        let mut buf = "LS: ".to_owned();
+
+        if busy {
+            // For RLS this can be "Build" or "Diagnostics" or "Indexing".
+            buf.push_str(
+                params
+                    .title
+                    .as_ref()
```
autozimu: You could use `.cloned()` to simply code here.

src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
...
+            // For RLS this is the crate name, present only if the progress isn't known.
+            if let Some(message) = params.message {
+                buf.push_str(&format!(" ({})", &message));
+            }
+            // For RLS this is the progress percentage, present only if the it's known.
+            if let Some(percentage) = params.percentage {
+                buf.push_str(&format!(" ({:.1}% done)", percentage));
+            }
+        } else {
+            buf.push_str("Idle");
```
autozimu: Whenever possible, `buf += ` can make code easier to read, IMO.

src/languageclient.rs
```diff
@@ -2074,6 +2074,50 @@ pub trait ILanguageClient: IVim {
...
+        }
+
+        // Sanitize before sending to vim.
+        buf.replace('\'', "''");
```
autozimu: Use the existing function `escape()` to do this.

autozimu: Plus, `.replace()` is not in-place modification.

src/languageclient.rs
YaLTeR: That's how it's done elsewhere; without it it gives an error that it expects a `Value` not an `Option`.

src/languageclient.rs
YaLTeR: This is necessary to turn `Option<String>` into an `Option<&str>` so I can `.unwrap_or(&'static str)` and prevent an extra allocation. An alternative would be `params.title.unwrap_or("Busy".into())` which would allocate for `"Busy"`.

YaLTeR:
Fixed or commented on the outlined issues.

src/languageclient.rs
autozimu: Ahh, I see. Apologies.

src/languageclient.rs
autozimu: I see.

autozimu:
Thanks.

PR #357, opened 2018-04-03
Remove __init__.py from rootMarkers

mfussenegger:
In python a common folder layout is something like:

    /
    /pkg_name
    /pkg_name/__init__.py
    /pkg_name/main.py
    /setup.py
    /setup.cfg

If `__init__py` is used as root marker the root will be `/pkg_name`.
This is for example problematic for
https://github.com/palantir/python-language-server because it picks up
certain settings from `setup.cfg` or `tox.ini` (among others)

(Maybe it would make sense to add `requirements.txt` or `Pipfile` instead)


autozimu:
> make sense to add requirements.txt or Pipfile instead

Could you add this? Thanks.

autozimu:
Thanks.

PR #399, opened 2018-04-24
Add `DiagnosticsList::Disabled` option

idanarye:
When picked, neither the quickfix list nor the loclist will be modified by LanguageClient-neovim.

The ability to set it to `v:null` was removed at [commit 5de7a52](https://github.com/autozimu/LanguageClient-neovim/commit/5de7a52a23c13097065f57487933dca0d30b0c02#diff-ba80c739c25b55f01f58c3724bcc93ddL295). I'm making it explicit to make sure it won't get accidentally removed again.

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -142,7 +142,7 @@ Valid options: "off" | "messages" | "verbose"
 List used to fill diagnostic messages.
 
 Default: "Quickfix"
-Valid options: "Quickfix" | "Location" | |v:null|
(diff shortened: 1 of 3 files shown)


autozimu:
Thanks for catching this and taking the effort to fix.

alok:
I updated to the latest version and tried setting `let g:LanguageClient_diagnosticsList   = "Disabled"` and got an error for it. The documentation indicates that `"Disabled"` should be a valid option as of this commit.

autozimu:
It's merged in but not included in release prebuilt binaries.

PR #407, opened 2018-04-29
Changing shebang

concatime:
`#!/usr/bin/env sh` ~> `#!/bin/sh`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/sh
 # Try install by
 #   - download binary
 #   - build with cargo


autozimu:
Why is this change necessary?

concatime:
Not really necessary, but we just follow the convention. Using the `/usr/bin/env` is a good habit when you need `bash`, but not useful if you rely only on `sh`. (For example, I'm using NixOS, and we have `/bin/sh` but not `/bin/bash`, and therefore, I use `/usr/bin/env bash`)

autozimu:
From this wikipedia section, https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability, `env` is still useful as POSIX doesn't enforce `/bin/sh` exists (and has the example of Solaris).

I don't think this change will have any immediate improvement. Thanks for contribution though.

PR #425, opened 2018-05-12
Use bash for install script

alok:
The script assumes bash only features anyway and the install instructions in the
README say to use bash.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 # Try install by
 #   - download binary
 #   - build with cargo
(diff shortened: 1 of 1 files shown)


autozimu:
Thanks for contributing.

I believe it's better to make the script compatible with `sh`, rather than hard depend on `bash`, as `bash` is not available by default in some systems like BSDs.

autozimu (changes requested):
Please check previous comment.

alok:
None of my systems lack `bash`, and it's not something I care about supporting, so I'll close the PR. I think in any case that a lot of BSD users will have installed bash for compatibility with other software that doesn't bother with strict POSIX compatibility.

Issue #435, opened 2018-05-16
Optionally do not stop server on exit vim


shkm:
In cases where a TCP socket is provided for a language server, the server needs to be started manually; this is fine, but I do find it unintuitive that exiting vim will still automatically stop the server.

My real-life use-case is in running Solargraph; the only way I'm aware to get it working with LanguageClient is to run the server myself.

As such, I propose a variable to control auto-stopping the language server. Ideally this could be configured on a per-language basis, but for now it'd be nice to just have the option.

autozimu:
Resolved with https://github.com/autozimu/LanguageClient-neovim/pull/436.

PR #436, opened 2018-05-16
#435 Introduce g:LanguageClient_autoStop

shkm:
So I'm not well-versed with vimscript enough to work on plugins, but this was such a minor change that I think this is fine. If not, consider it something of a proof of concept -- to save me from making a feature request and not really contributing anything to it.

Setting `g:LanguageClient_autoStop` to `0` will no longer stop the language server when vim exits. This is `1` by default (i.e. current behaviour).

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -123,28 +123,35 @@ associated filetype.
 
 Default: 1.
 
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks.

PR #438, opened 2018-05-18
Use TMPDIR for log path.

xu-cheng:
Per https://en.wikipedia.org/wiki/TMPDIR

TMPDIR is the canonical way to get the temporary path for POSIX.


autozimu:
Thanks.

Issue #428, opened 2018-05-13
有没有选项禁止 hover 时候弹出 preview window


skywind3000:
Preview window 极度干扰我的视线，打乱窗口布局，有没有办法hover时不显示preview window，只是再下面命令行 echo 一下即可？

mqudsi:
Google translation:

> Preview window is extremely disturbing my view, disrupt the window layout, there is no way to show the preview window hover, just the following command line echo can be?

skywind3000:
The google translation sucks.

Just request an option to disable preview window when calling LSP hover function. Opening a peview window accidentally will mess up my workspace and distract my focus.

Why this request is labeled as "help wanted" ?? I don't see any difficulty to disable the preview window, it is opened by LanguageClient-neovim itself.

KillTheMule:
> Why this request is labeled as "help wanted" ?? I don't see any difficulty to disable the preview window, it is opened by LanguageClient-neovim itself.

It's not alwas a question of difficulty, but also of priorities and total time spent. The maintainer has probably decided that he's open to merging a PR for that particular feature, but does not want to spend time on it by himself. Often, things are a tad more involved than is apparent on your first look, too.

skywind3000:
I suppose ui elements should be implemented in script, not in a native language like rust and it can be iterated rapidly. 

LanguageClient-nvim chose an opposite way and implement almost all the ui logics in rust. This will definitely slow down the iteration progress and become hard to add features in the future. 

And got less people who can help the development even if they want.

I suggest move the entire ui layer out of rust and keep the rust module as simple as a data service or message gateway between vimscript and Language Servers.

If so, more developers can get involved easily and help this plugin to become better.

skywind3000:
多谢 ！！

PR #443, opened 2018-05-23
Add g:LanguageClient_hoverPreview option

anthonysandrin:
Fixes: #428 

Add a new global option `g:LanguageClient_hoverPreview` with three valid values:
'Never': Never use preview window, always echo hover output
'Auto': Use preview window for hover entries longer than one line (default)
'Always': Always use preview window, never echo hover output

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -231,12 +231,12 @@ Valid options: number
 2.19 g:LanguageClient_hoverPreview             *g:LanguageClient_hoverPreview*
 
 Controls how hover output is displayed. Must be one of the following:
-    0 - Never use preview window, always echo hover output
-    1 - Use preview window for hover entries longer than one line (default)
-    2 - Always use preview window, never echo hover output
+    Never  - Never use preview window, always echo hover output
+    Auto   - Use preview window for hover entries longer than one line (default)
+    Always - Always use preview window, never echo hover output
 
-Default: 1
-Valid options: 0 | 1 | 2
+Default: "Auto"
+Valid options: "Never", "Auto", "Always"
 
 ==============================================================================
 3. Commands                                           *LanguageClientCommands*
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -228,6 +228,16 @@ before timing out.
 Default: 10
 Valid options: number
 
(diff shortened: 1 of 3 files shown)


autozimu (commented):
Thanks for contributing.

Please see my comments inline.

doc/LanguageClient.txt
```diff
@@ -228,6 +228,16 @@ before timing out.
...
 
+2.19 g:LanguageClient_hoverPreview             *g:LanguageClient_hoverPreview*
+
+Controls how hover output is displayed. Must be one of the following:
+    0 - Never use preview window, always echo hover output
+    1 - Use preview window for hover entries longer than one line (default)
+    2 - Always use preview window, never echo hover output
+
+Default: 1
+Valid options: 0 | 1 | 2
```
autozimu: For readability and extensibility, please use enum/string options, e.g., `never`, `auto`, `always`.

src/languageclient.rs
```diff
@@ -720,15 +720,18 @@ impl State {
             return Ok(result);
         }
 
+        let hover_preview: u8 = self.eval("get(g:, 'LanguageClient_hoverPreview', 1)")?;
```
autozimu: For efficiency, please check this setting inside function `sync_settings`.

src/languageclient.rs
```diff
@@ -720,15 +720,18 @@ impl State {
...
+
         let hover: Option<Hover> = serde_json::from_value(result.clone())?;
         if let Some(hover) = hover {
-            if hover.lines_len() <= 1 {
-                self.echo(hover.to_string())?;
-            } else {
+            if hover_preview == 2 || (hover_preview == 1 && hover.lines_len() > 1) {
                 self.preview(&hover.to_display())?;
+            } else {
+                self.echo(hover.to_string().lines().last().unwrap_or_default())?;
```
autozimu: Instead of showing only last line, use `echo_ellipsis` to show as much information as possible.

anthonysandrin: I agree that `echo_ellipsis` is a better solution so I switched to that. It's still not perfect; however, since the most relevant part of the hover output is usually the type signature and all of the language servers that I have tried (`rls` and `cquery` included) put the type signature at the end of the hover output.

Ideally I would be able to use the `signatureHelp` request to get this info, but not all servers support it 😞

autozimu:
Thanks. Neater than I would make it!

Issue #442, opened 2018-05-23
[Enhancement] Context menu


kristijanhusak:
LSP has a lot of options, and mapping all those or calling them manually is a bit annoying.
We could create a context menu that will give user a selection.
Something like this with a  built in `confirm` function:
![screenshot](https://i.imgur.com/kvRfMWe.png)

Or `inputlist`
![screenshot_inputlist](https://i.imgur.com/zjVnoML.png)

Maybe the second option is better because there's a lot of options, and first one goes horizontally, where the second one goes vertically.

If you are interested I'll create a PR.

autozimu:
Interesting idea! Please go for it.

DerWeh:
Sounds like a great idea. Maybe `LanguageClient-neovim` could just provide a `Denite` menu with its commands, as it already integrated with `Denite`.

KillTheMule:
That would restrict the feature to denite users, though.

Anrock:
How about **also** provide a Denite menu?

KillTheMule:
Sounds like a plan :)

kristijanhusak:
If we would support denite, then i guess we should support fzf too, like for everything else. This escalated quickly :D

Anrock:
Yeah. Denite and fzf support could be moved to separate issues, though.

kristijanhusak:
I'll try to create shared solution for all 3.

autozimu:
Merged in #446.

PR #446, opened 2018-05-24
Add context menu with all available actions. #442

kristijanhusak:
Here's implementation of context menu discussed in https://github.com/autozimu/LanguageClient-neovim/issues/442. I think i put all available actions in the dictionary. If i missed some, let me know.

By default, it uses native `inputlist()` with the numbered list, where you can select the action by number. Like a second screenshot in issue.

If fzf iz available, it wraps prompt in it, like documentSymbol for example.

For Denite users, `contextMenu` source is added.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -860,4 +860,50 @@ function! LanguageClient#cquery_vars(...) abort
     return call('LanguageClient#find_locations', ['$cquery/vars'] + a:000)
 endfunction
 
+function! LanguageClient_contextMenuItems() abort
+    return {
+                \ 'Code Action': 'LanguageClient#textDocument_codeAction',
+                \ 'Definition': 'LanguageClient#textDocument_definition',
+                \ 'Document Symbol': 'LanguageClient#textDocument_documentSymbol',
+                \ 'Formatting': 'LanguageClient#textDocument_formatting',
+                \ 'Hover': 'LanguageClient#textDocument_hover',
+                \ 'Implementation': 'LanguageClient#textDocument_implementation',
+                \ 'Range Formatting': 'LanguageClient#textDocument_rangeFormatting',
+                \ 'References': 'LanguageClient#textDocument_references',
+                \ 'Rename': 'LanguageClient#textDocument_rename',
+                \ 'Signature Help': 'LanguageClient#textDocument_signatureHelp',
+                \ 'Type Definition': 'LanguageClient#textDocument_typeDefinition',
+                \ 'Workspace Symbol': 'LanguageClient#workspace_symbol',
+                \ }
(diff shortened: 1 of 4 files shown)


autoload/LanguageClient.vim
```diff
@@ -860,4 +860,50 @@ function! LanguageClient#cquery_vars(...) abort
...
+  let l:options = keys(LanguageClient_contextMenuItems())
+
+  if get(g:, 'loaded_fzf') && get(g:, 'LanguageClient_FzfContextMenu', 1)
+    return fzf#run(fzf#wrap({
+    \ 'source': l:options,
+    \ 'sink': function('LanguageClient_handleContextMenuItem'),
+    \ }))
+  endif
+
+  let l:selections = map(copy(l:options), { key, val -> printf('%d) %s', key + 1, val ) })
```
autozimu: `copy()` is not necessary as far as I can see.

autoload/LanguageClient.vim
```diff
@@ -860,4 +860,50 @@ function! LanguageClient#cquery_vars(...) abort
     return call('LanguageClient#find_locations', ['$cquery/vars'] + a:000)
 endfunction
 
+function! LanguageClient_contextMenuItems() abort
+  return {
```
autozimu: Please keep the indent consistent, i.e., 4 spaces.

doc/LanguageClient.txt
```diff
@@ -267,6 +267,20 @@ Signature: LanguageClient#Notify(method: String, params: Map | List)
...
+Signature: LanguageClient#contextMenu(...)
+
+Show list of all available actions.
+
+If optional dependency FZF is installed, actions will be displayed in a FZF
+prompt, selecting one of the action will then call the action's function.
+
+To skip FZF prompt even if FZF is installed and use native numbered list,
+add this variable to vimrc:
+`let g:LanguageClient_FzfContextMenu = 0`
```
autozimu: Please keep option names consistent, i.e., `..._fzfContextMenu`. Also, please add doc to this option.

autozimu (commented):
Overall looks great!

Please see my inline comments.

autoload/LanguageClient.vim
kristijanhusak: It's needed because i'm using `l:options` at the end of the function, and map overrides it's content with the numbered list.

autoload/LanguageClient.vim
kristijanhusak: Fixed.

doc/LanguageClient.txt
kristijanhusak: Fixed.

autozimu:
Thanks.

PR #455, opened 2018-06-02
Don’t open hover preview window on empty output

wchargin:
Summary:
This fixes a bug in #443. With the default “auto” setting, the preview
window is supposed to open “for hover entries longer than one line”, but
in fact it opens for hover output with line count not exactly 1; that
is, it opens on empty output as well.

After this patch, empty hover output will result in an empty string
being echoed.

Test Plan:
With the default setting for this variable, trigger hover text with line
counts zero, one, and more than one, and note that only the last
category opens the preview window.

wchargin-branch: hover-no-preview-empty


autozimu:
Thanks.

PR #456, opened 2018-06-03
Fix match in completion

Synray:
Some language servers send non-keyword characters like ';',
so the old pattern doesn't match. Which caused situations like:
```c
fo# -> foo(#);

foo(ba#); -> foo(babar#);
```

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -775,7 +775,14 @@ function! LanguageClient#complete(findstart, base) abort
         let l:line = getline('.')
         let l:cursor = LSP#character()
         let l:input = l:line[:l:cursor]
-        let l:start = match(l:input, '\%(.\{-}[^\k]\{-}\)*\zs\k*\ze.\=$')
+
+        " Check for no identifier chars, but cursor is right after a non identifier char
+        " ie completing when the cursor is after the '.' in 'foo_class.'
+        if l:input[l:cursor-1:l:cursor-1] !~# "\k"
+            let l:start = l:cursor
+        else
+            let l:start = match(l:input, '\k*\ze[^\k]\=$')
+        endif
         return l:start
     else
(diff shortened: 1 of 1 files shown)


Synray:
Whoops, that's not actually correct. WIP

Synray:
Working correctly now.

autozimu:
I don't get what the problem is PR is trying to solve.

Could you give examples comparing the before and after behaviours?

Synray:
The problem I had was that if I had my cursor behind a `;` or any non-identifier character, and tried to complete anything, then any text I had entered to narrow the completion result wouldn't be removed.

The reason is because some language servers add the the `;` response. Two language servers that I use for C, cquery or ccls, did this.

#### Most frustrating example:
First I have `foo` which is assigned to `really_long_var_name`.
```c
int foo = really_long_var_name;
```
Now I want to change the variable's assignment to something else: `a_different_var_name`.
If I `ciw` at `really_long_var_name`, then type part of `a_different_var_name`, then what I had typed before completing is not removed like it should be.
So this: (cursor is at #)
```c
int foo = a_different#;
```
becomes this:
```c
int foo = a_differenta_different_var_name#;
```
It also happens when changing function arguments, which is what I was trying to show in the OP. For example trying to change insert `other_arg` in the middle of the function call:
```c
int foo = function1(arg1, other#, arg3);
```
becomes
```c
int foo = function1(arg1, otherother_arg#, arg3);
```

Synray:
I've just found another flaw in the regex, so if the commit looks good otherwise, please wait for me to fix it.

Synray:
That simplifies the regex.

Synray:
ARGH, still bugged.

Synray:
That bug was caused because `\k` apparently includes `.`. Is there any reason to avoid using `\W` at the end?

Synray:
Thanks, works now :)

PR #482, opened 2018-06-21
Add go LanguageServer to example, update function names in example mapping

balta2ar:


autozimu:
The example language server section is just example. It's not intended to list all and every language server.

Function names with `#` are lazy loaded. The ones without `#` are supposed to be replace by their `#` counterparts and eventually get deprecated, removed.

balta2ar:
I tried names with `#` and there were errors. Replacing with `_` worked.

PR #486, opened 2018-06-23
Change preview buffer name to avoid conflicts with fugitive

bennyyip:
When LanguageClient-neovim opens a preview buffer, my Vim echoes the following error message:
```
'Vim(let):E94: No matching buffer for /LanguageClient'
```
This is because [fugitive](https://github.com/tpope/vim-fugitive) does something nasty to buffers whose name is started with `//`. The code of fugitive is a little bit complex, I don't want to spend time on it. 

As many plugins surround temp buffer with "__", I change it from `//LanguageClient` to `__LanguageClient__`.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -618,7 +618,7 @@ impl State {
     where
         S: AsRef<str> + Serialize,
     {
-        let bufname = "//LanguageClient";
+        let bufname = "__LanguageClient__";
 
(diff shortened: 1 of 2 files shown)


autozimu:
Please fix the test.

bennyyip:
Fixed.

autozimu:
Thanks.

PR #488, opened 2018-06-26
Add jdt-language-server to Dockerfile and vimrc

Ferenc-:
A prerequisite for testing on Java data in container,
and for having a base configuration
for issue reproduction and verification.


autozimu:
I don't see strong benefits of having this inside the CI docker image when the language server is not used at all in the CI tests, especially considering the size of JDK.

Better would be creating separate docker images for manual testing and reproduction.

PR #490, opened 2018-06-26
Add cross compilation unit dependency for Java sample

Ferenc-:
Needed if/when there is an issue of LS not building dependencies


autozimu:
Closing as same reason of #488.

Issue #302, opened 2018-01-29
How to disable gutter?


ghost:
why? I think `texthl` has enough conspicuous.
I try to set the value of `signText` to an empty string, but there are still gutter:

```vim
let g:LanguageClient_diagnosticsDisplay = {
  \1: {'name': 'Error', 'texthl': 'ALEError', 'signText': '', 'signTexthl': 'ALEErrorSign',},
  \2: {'name': 'Warning', 'texthl': 'ALEWarning', 'signText': '', 'signTexthl': 'ALEWarningSign',},
  \3: {'name': 'Information', 'texthl': 'ALEInfo', 'signText': '', 'signTexthl': 'ALEInfoSign',},
  \4: {'name': 'Hint', 'texthl': 'ALEInfo', 'signText': '', 'signTexthl': 'ALEInfoSign',},
\}
```

So is there any option to disable gutter?
Thanks.

autozimu:
No, but happy to take in PR.

dcosson:
By disable gutter, do you also mean disable error checking on save altogether? That's a feature I'm looking for that i haven't found.

Now that ALE also has support for LSP, I'd like to stick with ALE for error & syntax checking (mainly because I'm already running some non-LSP linters through ale and it makes sense to me for all error checking & highlighting to be handled by one plugin for consistency. Plus ALE has some nice helpers like making the current error count available to display in powerline, and being able to jump to the next error which I don't think LanguageClient has).

But I still want to use LanguageClient when possible for things like showing the type under the cursor, jump to definition, refactoring, and other non-error-displaying LSP use cases.

Is this possible?

autozimu:
I think you need check this setting https://github.com/autozimu/LanguageClient-neovim/blob/7cf0f288ce80d633f3f68a6476260e0dd48f6bc0/doc/LanguageClient.txt#L147

Just to be clear, when LanguageClient is handling diagnostics (default behaviour), it will update quickfix/location list, so jumping to next error etc is supported.

matthewgrossman:
Not sure if this is a related issue;

 I'm in a similar boat to some users where I'm already setup w/ ALE as my linter/fixer but still want to use a Language Server for jump-to/completion/etc. I thought I could accomplish this with
```
let g:LanguageClient_diagnosticsEnable = 0
```
[...]

PR #492, opened 2018-06-30
Add an option to control the number of diagnostics signs

blyoa:
I added `g:LanguageClient_diagnosticsMaxSigns` option to control the number of diagnostics signs.

This option will solve the issue #302 by setting this option to `0`.


autozimu:
What is the use case for this change except to disable signs?

blyoa:
The use case is to keep neovim/vim snappy under many signs.  (According to [this gitgutter's doc](https://github.com/airblade/vim-gitgutter/blob/master/doc/gitgutter.txt#L317), many signs make neovim/vim slow.)

I decided the behavior of this option using [this option](https://github.com/airblade/vim-gitgutter/blob/master/doc/gitgutter.txt#L317) of gitgutter plugin and [this option](https://github.com/w0rp/ale/blob/ee81351a6393a6fef1b3556aebfdf522421018f3/doc/ale.txt#L1301) of ale plugin as references.

If the option should work more directly, I will update the pull-request; for example, I will change the option to `g:LanguageClient_diagnosticsDisplayEnable`.

autozimu:
I'm aware of the slowness of signs updating. That's why I've implemented an optimized diff-based approach to update the gutter signs. https://github.com/autozimu/LanguageClient-neovim/blob/92b047bea8e47cada801da9bfe1f3aa89eaa5775/src/utils.rs#L257-L283

With that said, this limit might be useful in some specific use cases.

Please see comments in line.

doc/LanguageClient.txt
```diff
@@ -105,6 +105,13 @@ Default: >
         },
     }
 
+2.3 g:LanguageClient_diagnosticsMaxSigns  *g:LanguageClient_diagnosticsMaxSigns*
```
autozimu: Sorry for the pedantic, but please update the name to g:LanguageClient_diagnosticsSignsMax

src/types.rs
```diff
@@ -120,6 +120,7 @@ pub struct State {
     pub diagnosticsEnable: bool,
     pub diagnosticsList: DiagnosticsList,
     pub diagnosticsDisplay: HashMap<u64, DiagnosticsDisplay>,
+    pub diagnosticsMaxSigns: i64,
```
autozimu: Please use Option for this setting, i.e., `Option<u64>`. When the option is not specified, it should be none.

src/languageclient.rs
```diff
@@ -117,6 +117,10 @@ impl State {
                 "has('nvim')",
             ].as_ref(),
         )?;
+
+        let (diagnosticsMaxSigns,): (i64,) =
```
autozimu: Please merge this with the previous lines. Every `eval` is an RPC call and thus has overhead.

autozimu (changes requested):

doc/LanguageClient.txt
blyoa: Thank you for the comments. I am happy to receiving any comments.
I updated the name..

src/types.rs
blyoa: I updated the type to `Option<u64>`.

src/languageclient.rs
blyoa: The size of tuple that `serde_json::from_value<T>` can interpret was up to 16, and I could not merge the variable with previous `let` lines (because already 16 variables exist in the `let` lines).
I moved the variables into a struct and assigned `serde_json::from_value<T>` as the struct. Now, the overhead was fixed.

Ref:
1. `serde_json::from_value<T>` calls `T::deserialize`: https://docs.serde.rs/src/serde_json/value/mod.rs.html#1121-1126
2. The defined tuple size of `T` is up to 16: https://docs.serde.rs/serde/trait.Deserialize.html

blyoa:
Thank you for the comments. I updated the pull-request.

src/languageclient.rs
autozimu: Ahh, sorry, I didn't know about this limitation.

Please use the original separate RPC call. (The current implementation obviously misses some properties when syncing to global `state`.)

blyoa: Thank you for the review. I updated the commit.

> The current implementation obviously misses some properties when syncing to global state.

I could not find what properties was missed. Could you pointed out some of the properties?
(This is my first rust program, and I'm sorry if I made an easy mistake.)
The previous commit is [this](https://github.com/blyoa/LanguageClient-neovim/commit/4ad3f3850b648992c5bdac53ebe10a6df0784092).

autozimu: My bad. I missed that some of the settings are copied from the intermediate `vimSettings` to local variables.

I noticed you updated the PR. Should be good to go.

autozimu:
Thanks.

PR #493, opened 2018-07-01
Fix FreeBSD detection

herrbischoff:
On FreeBSD, the `uname -sm` command gives the following output for 64-bit Intel processors:

```
$ uname -sm
FreeBSD amd64
```


autozimu:
Thanks.

PR #503, opened 2018-07-08
Add diagnostic highlighting to vim

lukel97:
Use `matchaddpos` to add highlight groups to diagnostics in plain old vim

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -441,6 +441,11 @@ impl State {
                 )?;
             }
         } else {
+            // Clear old highlights.
+            while let Some(match_id) = self.update(|state| Ok(state.highlight_match_ids.pop()))? {
+                self.call(None, "matchdelete", json!([match_id]))?;
+            }
+
             // Group diagnostics by severity so we can highlight them
             // in a single call.
             let mut match_groups: HashMap<_, Vec<_>> = HashMap::new();
@@ -454,6 +459,8 @@ impl State {
                     .or_insert_with(Vec::new)
                     .push(dn);
             }
+
+            let mut new_match_ids = Vec::new();
 
             for (severity, dns) in match_groups {
                 let hl_group = diagnosticsDisplay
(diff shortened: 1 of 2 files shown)


src/languageclient.rs
```diff
@@ -395,47 +395,105 @@ impl State {
...
+                        source,
+                        hl_group,
+                        dn.range.start.line,
+                        dn.range.start.character,
+                        dn.range.end.character,
+                    ]),
+                )?;
+            }
+        } else {
+            self.call::<_, u8>(None, "clearmatches", json!([]))?;
```
autozimu: Is this going to conflict with other plugins?

It's OK as a first step to support diagnostics highlight for vim.

src/languageclient.rs
```diff
@@ -395,47 +395,105 @@ impl State {
...
+                            // for each match.
+                            vec![[
+                                dn.range.start.line + 1,
+                                dn.range.start.character + 1,
+                                length,
+                            ]]
+                        } else {
+                            let mut middleLines: Vec<_> = (dn.range.start.line + 1
+                                ..dn.range.end.line)
+                                .map(|l| [l + 1, 0, 0])
```
autozimu: According to the doc, when trying to highlight the whole line, 2rd and 3rd item can be omitted. Plus, I'm not 100% sure if `0, 0` will be interpreted as whole line by vim.

src/languageclient.rs
```diff
@@ -395,47 +395,105 @@ impl State {
...
+                                .collect();
+                            let startLine = [
+                                dn.range.start.line + 1,
+                                dn.range.start.character + 1,
+                                999_999, //Clear to the end of the line
+                            ];
+                            let endLine = [dn.range.end.line + 1, 1, dn.range.end.character + 1];
```
autozimu: 3rd item should be length, not index.

autozimu (changes requested):

src/languageclient.rs
lukel97: `matchaddpos` returns an id that you can call `matchdelete` with to clear only that match. Would it be worthwhile to look at storing these ids in between calls to prevent conflicts?

autozimu: It's up to you if you want to include that within this PR.

I'm OK with current implementation.

src/languageclient.rs
autozimu: Sorry, my mistake. Should be correct.

src/languageclient.rs
lukel97: From doing a quick GitHub code search, I couldn't find any major plugins using `matchaddpos` but I've updated it to be a good citizen anyway

autozimu:
Thanks.

PR #506, opened 2018-07-09
Force vim to redraw when moving cursor

lukel97:
Currently on Macvim moving to a document symbol or reference requires the screen to be refreshed by scrolling or ctrl-L. Redrawing the screen fixes this. Does this happen on Neovim? If not then maybe there should be a check for just plain vim

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -463,14 +463,14 @@ impl State {
                     .ok_or_else(|| err_msg("Failed to get display"))?
                     .texthl
                     .clone();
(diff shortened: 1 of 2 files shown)


autozimu:
Are you running latest version of macvim?

I cannot reproduce the described issue.

lukel97:
When running `mvim --version`:
`VIM - Vi IMproved 8.1 (2018 May 17, compiled Jun 30 2018 01:47:01)
macOS version`
I will take a look at my setup and see if I'm doing anything funny that might be causing it

Issue #491, opened 2018-06-26
completion text not inserted correctly for haxe


jdonaldson:
## Summary

I'm using the haxe completion server, which will generate fields and top level completions with the following json :  https://gist.github.com/jdonaldson/6f3542412c7569fedadc5d41be4ddf8f

I've been following the completion changes from the code that addressed https://github.com/autozimu/LanguageClient-neovim/issues/478.

Here's what's working for me:

1. I can configure the g:LanguageClient_completionPreferTextEdit option to get the new behavior.
2. completions are using the correct display label, and using the correct "word" for insertion.
3. completions are triggering the behavior for "additionalText".

However, I'm noticing that text is not always inserted correctly.  E.g  for this example (cursor at "|"):

```haxe
class Main {
  static function main(){
    "hi".|
  }
}
```

The completions will trigger successfully for string methods.  However, once a method is chosen (e.g. `charAt`), the result is this:

```haxe
|class Main {
    static function main() {
    }
}
charAt

```

It looks like the original instance "hi" was deleted.  Then, the chosen completion was inserted at the end of the document, and the cursor was moved to the beginning.

The positions/ranges for the completion look correct to me.  I was noticing if I saved the document before triggering the completion, it would behave correctly:

```haxe
class Main {
    static function main() {
        "hi".charAt|
    }
}
```

A similar behavior happens for top level completions (which trigger imports of classes).   E.g. this is what the importing the class "Async" looks like:

```haxe

import lua.lib.luv.Async;

class Main {|
    static function main() {


    }
}
Async
```

Saving the file first does not help in this case, and this is the result:

```haxe
import lua.lib.luv.Async;

class Main {|
    static function main() {
        AsyncAsync 


    }
}
```

Issue #478, opened 2018-06-19
Support "newText" in completion items


jdonaldson:
## Support "newText" in completion items

The Haxe languageserver uses newText instead of insertText for declaring code content to include in completions: https://github.com/vshaxe/haxe-languageserver/blob/d7799fa0570b91f10ec272ee4e7287f0224acf09/src/haxeLanguageServer/features/completion/CompletionFeature.hx#L275

This plugin uses insertText (insert_text) with a fall back to the label: https://github.com/autozimu/LanguageClient-neovim/blob/next/src/types.rs#L934

It looks like insertText usage is deprecated according to Microsoft : https://github.com/Microsoft/language-server-protocol/issues/264

What's the plan for handling insertText deprecation?  My 2 cents would be to:

1. Check for new_text here in this plugin, and use it if present.
2. If it is not present, check for insert_text, and use it (while also displaying a warning message about deprecation)
3. If neither are present, use the label.

I would give you a pull request, but I'm not very familiar with rust.  I can build the project, but it doesn't appear to work in my vim session.

autozimu:
It's not so obvious as it looks.

With the new insertText, the `newText` might not be starting from completion starting point. Say
```
const varname;
sort(var<cursor>
```
With `insertText`, it could be that the `newText` property is `sort(varname)` with range start to be 0.

I've actually started the work to support this. Check https://github.com/autozimu/LanguageClient-neovim/blob/1004edf2542168f1a15f75d9078590b4f59dae89/src/types.rs#L958 

Just a few issues remaining before I can release it.

jdonaldson:
Good point.  

Thanks for putting the thought into it.  I'll look forward to the update.

PR #509, opened 2018-07-11
fix positioning of edits when applying text edits

jdonaldson:
Here's a sketch of some attempts to fix the positioning problem for the haxe language server.
I have another branch over on the haxe language server itself with some complimentary changes : https://github.com/vshaxe/haxe-languageserver/pull/35#pullrequestreview-129359549

Please bear with me, this is my first time working with rust!

There's actually a few problems at play here.  Ocassionally I'm not sure if fixes are more appropriate in the nvim client, or over on the haxe langserver side.  I'll go through and annotate the sections below. 

This pull request addresses the issues here in a more direct fashion: 
https://github.com/autozimu/LanguageClient-neovim/issues/478
https://github.com/autozimu/LanguageClient-neovim/issues/491

cc @roxma

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -2202,16 +2202,42 @@ impl State {
         let user_data: VimCompleteItemUserData = serde_json::from_str(&user_data)?;
 
         let mut edits = vec![];
+        let mut cur_line = line + 1;
+        let mut cur_char = character + 1;
         if let Some(edit) = user_data.text_edit {
-            self.command("undo")?;
-            edits.push(edit.clone());
+            let word_len = completed_item.word.len();
+            let start : usize  = &(character as usize) - word_len;
+            let end : usize = character as usize;
+            let mut current_line: String = self.call(None, "getline", json!([line+1]))?;
+            current_line.replace_range(start..end, "");
+
+            if self.call::<_, i64>(None, "setline", json!([line+1, current_line]))? != 0 {
+                bail!("Failed to set buffer content!");
+            }
+            let mut edit_clone = edit.clone();
+            edit_clone.range.start.character = start as u64;
+            edit_clone.range.end.character = start as u64;
+
+            edits.push(edit_clone);
         };
         if let Some(aedits) = user_data.additional_text_edits {
+             for edit in aedits.iter()  {
+                if edit.range.start.line.to_usize()? < cur_line.to_usize()? {
+                    cur_line += edit.new_text.lines().count() as u64;
+                }
+                if edit.range.start.line.to_usize()? == cur_line.to_usize()? {
+                    if edit.range.end.character.to_usize()? < cur_char.to_usize()?{
+                        cur_char += edit.new_text.len() as u64;
+                    }
+                }
+            }
             edits.extend(aedits.clone());
         };
 
+
+
         self.apply_TextEdits(filename, &edits)?;
-        self.cursor(line + 1, character + 1)
+        self.cursor(cur_line, cur_char+1)
     }
 
     pub fn languageClient_FZFSinkLocation(&mut self, params: &Option<Params>) -> Result<()> {
--- a/src/types.rs
+++ b/src/types.rs
@@ -953,10 +953,15 @@ pub fn to_vim_complete_item(
     preferTextEdit: bool,
 ) -> Result<VimCompleteItem> {
     let abbr = lspitem.label.clone();
+    let new_text = match lspitem.text_edit.clone() {
+        Some(text_edit) => text_edit.new_text,
+        None => lspitem.label.clone()
+    };
+
     let word = lspitem
         .insert_text
         .clone()
-        .unwrap_or_else(|| lspitem.label.clone());
+        .unwrap_or_else(|| new_text);
 
     let mut user_data = VimCompleteItemUserData::new();
 


src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
         let user_data: VimCompleteItemUserData = serde_json::from_str(&user_data)?;
 
         let mut edits = vec![];
+        let mut cur_line = line + 1;
+        let mut cur_char = character + 1;
```
jdonaldson, line 2206: Here I'm setting up some mutable variables for the line/char index.  These can change due to additional text inserts, etc.

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
...
+            current_line.replace_range(start..end, "");
+
+            if self.call::<_, i64>(None, "setline", json!([line+1, current_line]))? != 0 {
+                bail!("Failed to set buffer content!");
+            }
+            let mut edit_clone = edit.clone();
+            edit_clone.range.start.character = start as u64;
+            edit_clone.range.end.character = start as u64;
+
+            edits.push(edit_clone);
```
jdonaldson, line 2221: "undo" in this situation causes problems, since triggering an autocomplete does not break the current insert state.  The undo operation could nuke leading whitespace and newlines, and throw the completion position off.

Instead, we calculate the start/end boundaries of the inserted word, then we tell vim to remove that from the given line.

Also, I'm noticing that the character offsets given by the language server can be wrong here.  This is because completion clients will cache the results, and then filter them on subsequent keypresses.  Here I'm just hard coding the range of the edit to be the start position of the originally inserted completion.  One alternative might involve manually inserting a break into the insert operation `ctrl-g u`: http://vimdoc.sourceforge.net/htmldoc/insert.html#i_CTRL-G_u

Another question I had is whether we should remove the "normal" completion text at all.  I realize text edits are more flexible, but it makes the code here more complicated to support that, and I'm not aware of a situation where you would need the extra complexity.

*Hopefully there's a better way of handling these issues, any ideas?*

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
...
+             for edit in aedits.iter()  {
+                if edit.range.start.line.to_usize()? < cur_line.to_usize()? {
+                    cur_line += edit.new_text.lines().count() as u64;
+                }
+                if edit.range.start.line.to_usize()? == cur_line.to_usize()? {
+                    if edit.range.end.character.to_usize()? < cur_char.to_usize()?{
+                        cur_char += edit.new_text.len() as u64;
+                    }
+                }
+            }
```
jdonaldson, line 2233: Adding additional text edits commonly inserts content at the beginning of the document.  This can throw off the cursor position below it.  We need to sweep through these edits and calculate the line and character offsets they cause.

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
...
             edits.extend(aedits.clone());
         };
 
+
+
         self.apply_TextEdits(filename, &edits)?;
-        self.cursor(line + 1, character + 1)
+        self.cursor(cur_line, cur_char+1)
```
jdonaldson, line 2240: Updating the cursor with the corrected values.

src/types.rs
```diff
@@ -953,10 +953,15 @@ pub fn to_vim_complete_item(
     preferTextEdit: bool,
 ) -> Result<VimCompleteItem> {
     let abbr = lspitem.label.clone();
+    let new_text = match lspitem.text_edit.clone() {
+        Some(text_edit) => text_edit.new_text,
+        None => lspitem.label.clone()
+    };
+
```
jdonaldson, line 960: The haxe language server uses the new "new_text" field for specifying the text.  This doesn't really matter much, since the text edit process in languageclient.rs will fix things.  I still think it's a good idea to have the right config for abbr/word in the base completion behavior though.  *I also would prefer new_text to be the preferred default, and insert_text to be the fallback, if that's ok I can make that change as well*.

src/languageclient.rs
jdonaldson: FWIW, this won't handle text edits that delete text.

src/languageclient.rs
autozimu: I don't think at the moment there is any better way of handling this awkward situation, unless it is supported from vim side https://github.com/neovim/neovim/issues/8334

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
...
+            let end : usize = character as usize;
+            let mut current_line: String = self.call(None, "getline", json!([line+1]))?;
+            current_line.replace_range(start..end, "");
+
+            if self.call::<_, i64>(None, "setline", json!([line+1, current_line]))? != 0 {
+                bail!("Failed to set buffer content!");
+            }
+            let mut edit_clone = edit.clone();
+            edit_clone.range.start.character = start as u64;
+            edit_clone.range.end.character = start as u64;
```
autozimu, line 2219: This is really wired. Is this necessary or even working?

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
         let user_data: VimCompleteItemUserData = serde_json::from_str(&user_data)?;
 
         let mut edits = vec![];
+        let mut cur_line = line + 1;
+        let mut cur_char = character + 1;
         if let Some(edit) = user_data.text_edit {
-            self.command("undo")?;
-            edits.push(edit.clone());
+            let word_len = completed_item.word.len();
+            let start : usize  = &(character as usize) - word_len;
```
autozimu, line 2209: `&` might be unnecessary.

src/languageclient.rs
```diff
@@ -2202,16 +2202,42 @@ impl State {
...
+            }
+            let mut edit_clone = edit.clone();
+            edit_clone.range.start.character = start as u64;
+            edit_clone.range.end.character = start as u64;
+
+            edits.push(edit_clone);
         };
         if let Some(aedits) = user_data.additional_text_edits {
+             for edit in aedits.iter()  {
+                if edit.range.start.line.to_usize()? < cur_line.to_usize()? {
```
autozimu, line 2225: `.to_usize()?` should be unnecessary.

src/languageclient.rs
autozimu: This and following calculation are all 0-based. It might be better to keep 0-based until calling vim api.

src/types.rs
autozimu: `insert_text` is more accurate whenever it is available in situation that `new_text` didn't start from current completion starting point.

autozimu (commented):
Thanks a lot for working on this.

Please see my comments inline.

src/languageclient.rs
jdonaldson: yes for both.  What's going on here is:

1. Removing the word inserted by base autocomplete behavior.  (replacing the "undo" command)
2. Editing the start/end position of the ranges from the given completion response. (the completion response is cached by completion managers, and often does not have the correct position according to the current cursor position).

I talk more about them below.

src/types.rs
jdonaldson: I think it's ok to use deprecated insert_text when it's available, but you should check for new_text as well rather than immediately falling back to the label.  

Going further, when you get to the "completedone" phase, you can check to see if the completed word is the last (bottom-most/right-most) edit in the list of text edits to make.  If it is, then you don't need to redo it.  You can just apply the rest of the additionalTextEdits.

Does that make sense?

autozimu: I don't think you got my point here.

`TextEdit`'s `newText` is in a way not compatible with all existing vim completion inforstructure, because the `newText` might not be starting at expected completion starting point by vim.

For example,
```
let language<cursor>
```
where <cursor> is current cursor position.

With vim completion convention (also deoplete's), the returned completion items must all begin with `language`.

Whereas with LSP's `TextEdit`, the `newText` could be `let langugeclient`, with a range start equals to 0.

If in this case, we feed `newText` (`let languageclient`) as `word` to vim, vim completion will behave annoyingly of eating existing input, and in deoplete, it won't be displayed in completion list at all as it doesn't match existing input `language`.

On the other hand, `insertText` and `label` are almost always start with the correct completion starting point.

jdonaldson: You're right newText is not completely compatible with how vim does completion functions.  However:

1. Most of the time it is, and you can detect it (start/end position are equal to the position the completion was requested at).
2. If it's not, your applyTextEdit method can sort it out, as long as you delete (correctly) the text inserted by the completion engine.

It's true that some of the filtering might be off, at least for some completion engines.  However, that is really a problem for the completion engine, not your plugin.  FWIW deoplete looks like it does the fuzzy matching behavior I would like to have.

Personally, insertText is much simpler for me to deal with.  However, I don't want to try and argue to the haxe languageserver team to support an obsolete field.  FWIW, I've identified some issues on their end where they diverge from spec as well.  I don't personally love the spec, but I really think it's the only way to make sure our code all works together consistently.

jdonaldson: On a side note, I'll also mention that my rust skills are pretty bad.  I'd hope that you could understand what I'm trying to do and make cleaner/idiomatic changes.  E.g. if you see extraneous `to_usize()` methods, please assume I just didn't know what I was doing with low level language features (which is almost certainly true).

butterflyfish:
great job! I'm looking forward it's merged.

autozimu:
I believe a similar improvement has been merged through https://github.com/autozimu/LanguageClient-neovim/pull/858

Free free to reopen if I missed anything. Thanks for contributing!

cheuk-fung:
@autozimu I think #858 did not address the cursor position issue caused by newlines in `additional_text_edits`. More detailed description for this issue: #827

update: I submitted a pull request #961 to fix the cursor position.

PR #512, opened 2018-07-12
skip handleCompleteDone if edits is empty, fix ncm2/ncm2#16

roxma:
ncm2/ncm2#16

@HybridEidolon please test this PR

Note that you need to recompile LanguageClient-neovim and make sure the `bin/languageclient` is the latest build.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -2210,6 +2210,10 @@ impl State {
             edits.extend(aedits.clone());
         };
 
+        if edits.is_empty() {
+            return Ok(())
+        }
+
         self.apply_TextEdits(filename, &edits)?;
         self.cursor(line + 1, character + 1)
     }


autozimu:
I don't get it why this happens in the first place.

When returning completion items, there is a check before adding `user_data`, https://github.com/autozimu/LanguageClient-neovim/blob/c01b4ef60dc260739aeb94f7494f811c8ba348e2/src/types.rs#L996-L1000

And before calling `CompletionDone`, there is another https://github.com/autozimu/LanguageClient-neovim/blob/c01b4ef60dc260739aeb94f7494f811c8ba348e2/autoload/LanguageClient.vim#L742-L745

Is ncm2 are inserting extra data to `user_data`?

roxma:
ncm2 is using `user_data` field too. There're non lsp based plugins using user_data  for snippet, and match highlight.

Even if the user_data is missing when ncm2 receives the matches, ncm2 will automatically formalize it, with empty user_data, and some exart defaults, which make things easier for filtering, processing, without assuming that some data may be missing.

autozimu:
I see. Make sense.

HybridEidolon:
@roxma It does not appear to fix ncm2/ncm2#16. Sorry for the delay.

Issue #516, opened 2018-07-15
Cannot get this plugin working with typescript-language-server (this plugin does not allow commandline arguments)


dylan-chong:
> If you are reporting a bug, please read following FAQ first. If you are
requesting other types of issue, please ignore.

- Did you upgrade to latest plugin version? Yes
- Did you upgrade to/compile latest binary? Run shell command
  `bin/languageclient --version` to get its version number. `languageclient 0.1.99 05c4b93c58c524a83bd30b388ec7ff74888ebd49`
- (Neovim users only) Did you check output of `:checkhealth LanguageClient`? Yes
- Did you check [troubleshooting]? Yes

## Summary

With this configuration:

```
call plug#begin('~/.vim/plugged')

Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }

call plug#end()

augroup filetype_rust
    autocmd!
    autocmd BufReadPost *.rs setlocal filetype=rust
augroup END
augroup filetype_typescript
    autocmd!
    autocmd BufReadPost *.ts setlocal filetype=typescript
augroup END


" Always draw sign column. Prevent buffer moving when adding/deleting sign.
set signcolumn=yes

let g:LanguageClient_serverCommands = {
    \ 'rust': ['rustup', 'run', 'stable', 'rls'],
    \ 'typescript': ['typescript-language-server --stdio']
    \ }
let g:LanguageClient_loggingFile = '/tmp/LanguageClient.log'
let g:LanguageClient_loggingLevel = 'INFO'
let g:LanguageClient_serverStderr = '/tmp/LanguageServer.log'

```

I get this log output

```
14:00:28 WARN main src/languageclient.rs:2090 Failed to start language server automatically. No such file or directory (os error 2)
```

Presumably because 'typescript-language-server --stdio' gets escaped. 

## Reproduction
- neovim/vim version (`nvim --version` or `vim --version`): nvim 0.3
- This plugin version (`git rev-parse --short HEAD`):  Mentioned above
- This plugin's binary version (`bin/languageclient --version`): Mentioned above
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue. Refer to an example [here][min-init.vim]): The above
- Language server link and version: https://github.com/theia-ide/typescript-language-server/releases/tag/v0.1.11 (install globally through npm/yarn
- Steps to reproduce the issue from clean state
[...]

PR #518, opened 2018-07-15
Clarify that arguments can be passed to the server commands

dylan-chong:
Clarifies solution for #516

**Please squash commits on merge**


autozimu:
Thanks.

PR #544, opened 2018-08-03
Handle symlinks when switching to buffers.

rickyz:
In some setups, the editor and language server may access a file via
different paths due to symlinks.

In Vim, bufnr(path) will resolve symlinks when doing patch matches.
Unlike bufnr, :buffer [bufname] does not resolve symlinks.

Before this change, when the editor/language server have different
symlinked paths, attempting to go to an already-open file would result
in "No matching buffer" errors. Fix this by passing the buffer number to
:buffer instead.


autozimu:
Thanks.

PR #545, opened 2018-08-04
Echo some info when calling gotoDefinition

liuchengxu:
Although `CTRL-G` could also show the current file name, the cursor position and so on, it's nice to have these info without extra operation when calling gotoDefinition from LCN. I know vim-go also echoes some info in this case and I think that's very useful.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -953,6 +953,8 @@ impl State {
                     let loc = arr.get(0).ok_or_else(|| err_msg("Not found!"))?;
                     self.edit(&goto_cmd, loc.uri.filepath()?)?;
                     self.cursor(loc.range.start.line + 1, loc.range.start.character + 1)?;
+                    let cmd = format!("echohl Function | echom '[LC]: '.expand('%').' {}:{}' | echohl NONE", loc.range.start.line + 1, loc.range.start.character + 1);
+                    self.command(&cmd)?;
                 }
                 _ => self.display_locations(&arr)?,
             },
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -953,8 +953,8 @@ impl State {
                     let loc = arr.get(0).ok_or_else(|| err_msg("Not found!"))?;
                     self.edit(&goto_cmd, loc.uri.filepath()?)?;
                     self.cursor(loc.range.start.line + 1, loc.range.start.character + 1)?;
-                    let cmd = format!("echohl Function | echom '[LC]: '.expand('%').' {}:{}' | echohl NONE", loc.range.start.line + 1, loc.range.start.character + 1);
-                    self.command(&cmd)?;
+                    let cur_file: String = self.eval("expand('%')")?;
+                    self.echomsg_ellipsis(format!("[LC]: {} {}:{}", cur_file, loc.range.start.line + 1, loc.range.start.character + 1))?;
                 }
                 _ => self.display_locations(&arr)?,
             },


autozimu:
There're two concerns before merging this in.

1. If the message is a long one, will it trigger |hit-enter| to continue prompt?
2. This change only covers one case of GotoLocation. Are you planning to make the same behaviour for the other case, i.e., when there are multiple locations available?

liuchengxu:
1. Make sense. The long message concern could definitley be addressed in two ways: one is `s:Ellipsis()` in LCN, the other is [something like ale#cursor#TruncatedEcho()](https://github.com/w0rp/ale/blob/9674132933df3e5b126c30da33a26c9be3ef87c6/autoload/ale/cursor.vim#L12-L38), see `:h shortmess`. IMHO the latter one is more native. If allowed, I could open up another PR for this point.

2. Except for GotoDefinition, I personally rarely use the other GotoLocations. However, if you could give me a hand, I'd love to take a stab at covering more cases, although my knowledge about LCN and rust is very limited :).

autozimu:
ALE's workaround seems better as it saves complete messages to message history.

autozimu:
For multiple locations of Goto, it is handled at https://github.com/autozimu/LanguageClient-neovim/blob/e5c879db087e385c6a8e7ecf224b36077d916e54/src/languageclient.rs#L2302

liuchengxu:
I applied the same logic to the multiple locations of Goto you pointed out, but the user experience is seemingly not very good right now. For instance, when goto a location in another file, there is an obvious delay between echoing the info and moving the cursor to the right position, which results in echoed info disappeared before moving the cursor. I'll find some time to look into this.

autozimu:
It might because of conflicts between https://github.com/autozimu/LanguageClient-neovim/blob/e5c879db087e385c6a8e7ecf224b36077d916e54/src/languageclient.rs#L2142

Whenever cursor changes, it will try to display line diagnostics, which might override this echoed info.

liuchengxu:
1. I have used `echomsg_ellipsis()` to avoid the long messages. There is still room for improvement, e.g., `echohl Function`. 

2. The echo attempt of diagnostics seems to be blamed. Any clue to *fix* it? I have no idea by far.

src/languageclient.rs
```diff
@@ -953,6 +953,8 @@ impl State {
                     let loc = arr.get(0).ok_or_else(|| err_msg("Not found!"))?;
                     self.edit(&goto_cmd, loc.uri.filepath()?)?;
                     self.cursor(loc.range.start.line + 1, loc.range.start.character + 1)?;
+                    let cur_file: String = self.eval("expand('%')")?;
+                    self.echomsg_ellipsis(format!("[LC]: {} {}:{}", cur_file, loc.range.start.line + 1, loc.range.start.character + 1))?;
```
autozimu, line 957: GotoDefinition doesn't necessarily go the same file. This echoed filename should be the target file, right?

liuchengxu: Yes, the echoed filename is always from the target file. If the target is same with current file, then just echo current filename with cursor position changed alone.

autozimu: OK. Then it's not necessary to use `eval` to get the filename. It's better to get that from `loc` variable. Every `eval` is an rpc call and thus has larger overhead.

liuchengxu: Here is my revision:

```rust
self.echomsg_ellipsis(format!("[LC]: {} {}:{}", loc.uri.filepath().unwrap().display(), loc.range.start.line + 1, loc.range.start.character + 1))?;
```

However, it always shows the absolute path:

```
[LC]: /Users/xlc/.vim/plugged/LanguageClient-neovim/src/vim.rs 214:12
```

whereas I hope when I'm still inside the project, it's better to show the relative path:

```
[LC]: src/vim.rs 214:12
```

Is there another solution?

autozimu: I see. Make sense.

autozimu:
Thanks.

PR #549, opened 2018-08-05
Fixed FZFSinkCommand for commands with ": " in the title

Jesse-Bakker:
We only use the first two parts, so we only split in two parts to allow for titles with ": " in the title.


autozimu:
Thanks.

PR #552, opened 2018-08-07
Utilize shortmess option to truncate long messages

liuchengxu:


autozimu:
Maybe I did something stupid.

When I'm trying this out, long messages are not displayed at all.

Can you double check?

screen shot 2018-08-07 at 18 28 03
screen shot 2018-08-07 at 18 28 16

liuchengxu:
Hmm, I can reproduce the diagnostics messages are gone.



but I also believe `s:Truncate()` works as expected.



What's more, GotoDefinition is good:



```vim
call plug#begin('~/.vim/plugged')

    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }

call plug#end()

augroup filetype_rust
    autocmd!
    autocmd BufReadPost *.rs setlocal filetype=rust
augroup END

" Always draw sign column. Prevent buffer moving when adding/deleting sign.
set signcolumn=yes

    \ 'rust': ['rustup', 'run', 'nightly', 'rls'],
    \ }
let $RUST_BACKTRACE = 1
let g:LanguageClient_loggingFile = '/tmp/LanguageClient.log'
let g:LanguageClient_loggingLevel = 'INFO'
let g:LanguageClient_serverStderr = '/tmp/LanguageServer.log'

nnoremap gd :call LanguageClient#textDocument_definition()<CR>

set noswapfile
```

So, the problem is perhaps related to how/when the linter messages will echo? I haven't found the root cause :(.

autozimu:
Yes, it's probably caused by something else. But still we needs to figure out why.

autozimu:
I think the problem is because with `shortmess+=T`, `echo someLongMessage` won't be displayed.

Can you revert the changes made to `s:EchoEllipsis`?

autoload/LanguageClient.vim
```diff
@@ -6,30 +6,36 @@ function! s:Echo(message) abort
...
-        let maxlen -= 18
-    endif
-    if len(a:message) < l:maxlen
-        let l:message = a:message
-    else
-        let l:message = a:message[:l:maxlen - 3] . '...'
-    endif
-    return l:message
+" Credit: ALE, snippets from ale#cursor#TruncatedEcho()
+function! s:Truncate(message, cmd) abort
```
autozimu: Can you add also the replacing of line endings?

liuchengxu: I don't understand very well. What does `the replacing of line endings` mean? Anyway, I have pushed a commit to revert the `s:EchoEllipsis()` related changes. Correct me if anything is not suitable.

autozimu: I meant replacing line endings and tabs from original ALE implementation,

https://github.com/w0rp/ale/blob/9674132933df3e5b126c30da33a26c9be3ef87c6/autoload/ale/cursor.vim#L17

autoload/LanguageClient.vim
```diff
@@ -27,9 +27,31 @@ function! s:EchoEllipsis(message) abort
     echo s:Ellipsis(a:message)
 endfunction
 
+" Credit: ALE, snippets from ale#cursor#TruncatedEcho()
+function! s:Truncate(message) abort
```
autozimu: Please update this function name or merge the function body into EchomsgEllipsis.

autoload/LanguageClient.vim
liuchengxu: Done.

autoload/LanguageClient.vim
liuchengxu: Done.

autozimu:
Thanks!

PR #555, opened 2018-08-10
Fix ncm2/ncm2#58 and ncm2/ncm2#59

roxma:
- handleTextChanged in TextChangedP
- force  `is_incomplete=true` when there's something wrong to invalidate ncm2's cache

https://github.com/ncm2/ncm2/issues/58

https://github.com/ncm2/ncm2/issues/59

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -2430,23 +2430,32 @@ impl State {
         let line = ctx.lnum - 1;
         let character = ctx.ccol - 1;
 
-        let result = self.textDocument_completion(
-            &json!({
+        let mut result: Result<Value> = Ok(Value::Null);
+        let completion = self.textDocument_completion(&json!({
                 "buftype": "",
                 "languageId": ctx.filetype,
                 "filename": filename,
                 "line": line,
                 "character": character,
-                "handle": false,
-            }).to_params()?,
-        )?;
-        let result: Option<CompletionResponse> = serde_json::from_value(result)?;
-        let result = result.unwrap_or_else(|| CompletionResponse::Array(vec![]));
-        let is_incomplete = match result {
+                "handle": false}).to_params()?)
+            .unwrap_or_else(|err| {
+                result = Err(err);
+                Value::Null
+            });
+        let completion: CompletionResponse = serde_json::from_value(completion)
+            .unwrap_or_else(|err| {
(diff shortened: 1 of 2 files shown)


osleg:
This fixed the issue with unreliable autocomplete

src/languageclient.rs
```diff
@@ -2384,9 +2384,12 @@ impl State {
                 "line": line,
                 "character": character,
                 "handle": false,
-            }).to_params()?)?;
-        let result: Option<CompletionResponse> = serde_json::from_value(result)?;
-        let result = result.unwrap_or_else(|| CompletionResponse::Array(vec![]));
+            }).to_params()?).unwrap_or(Value::Null);
```
autozimu: This and following line could be simplified with `.and_then`.

```
let result = completion()
    .and_then(|response| serde_json::from_value(response))
    .unwrap_or_else(|| default value);
```

Another concern is that with this change, the real cause might be covered up and thus never got properly fixed.

autozimu: Especially under https://github.com/ncm2/ncm2/issues/59, it mentions 
> golang's language server will response an error. It looks like the language client send an invalid position to server

roxma: @autozimu 

> Another concern is that with this change, the real cause might be covered up and thus never got properly fixed.

This is fixed in https://github.com/autozimu/LanguageClient-neovim/pull/555/commits/f1c15bd40b4c4cb73ce92c277ffdca90c929c1d2

src/languageclient.rs
```diff
@@ -2430,23 +2430,32 @@ impl State {
...
-        let is_incomplete = match result {
+                "handle": false}).to_params()?)
+            .unwrap_or_else(|err| {
+                result = Err(err);
+                Value::Null
+            });
+        let completion: CompletionResponse = serde_json::from_value(completion)
+            .unwrap_or_else(|err| {
+                if result.is_ok() {
+                    result = Err(err.into());
```
autozimu, line 2448: Is this going to reassign result when it is ok?

If answer is no, then the result is reassigned to the second error caused by `from_value(null)` instead of the first error returned by language server.

autozimu: Anyway, I think I got the idea what you're trying to achieve. I can make some refinements some time today.

PR #561, opened 2018-08-11
Implement textDocument/documentHighlight

YaLTeR:
This request highlights usages of the symbol under the cursor.

Usage: position cursor on symbol (like hover), `:call LanguageClient_textDocument_documentHighlight()`.

The highlights are controlled with a new variable, `g:LanguageClient_documentHighlightDisplay`, in a similar fashion to the existing `g:LanguageClient_diagnosticsDisplay`. The default values are `LanguageClient{Text,Read,Write}` for text mentions, variable reads and writes, respectively.

Test vimrc:
```vim
call plug#begin('~/.local/share/nvim/plugged')

    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }

call plug#end()

augroup filetype_rust
    autocmd!
    autocmd BufReadPost *.rs setlocal filetype=rust
augroup END

    \ 'rust': ['rustup', 'run', 'stable', 'rls'],
    \ 'python': ['pyls'],
\ }

highlight link LanguageClientText Visual
highlight link LanguageClientRead Visual
highlight link LanguageClientWrite FoldColumn
```
Example (RLS, doesn't do read/write):
image
Example (pyls, does read/write):
image

Known issues:
- I clear the highlight on `CursorMoved` (because it's likely that the cursor was moved out of the symbol), but currently `CursorMoved` is only issued when the cursor moves up or down, not left or right.
  - Also due to clearing in `CursorMoved` I can't reliably call this from my own `CursorMoved` (since it can end up being called before LanguageClient's and the highlight will be immediately erased). Not sure how to handle this.
  - An alternative would be to not clear the highlights automatically and instead add a function to clear them.
- I'm not sure what the best way of adding vim support is.

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -281,6 +281,26 @@ neovim >= 0.3.0.
 Default: 0
 Valid options: 1 | 0
 
+2.22 g:LanguageClient_documentHighlightDisplay *g:LanguageClient_documentHighlightDisplay*
+
+Control how document highlights are displayed.
+
+Default: >
+    {
+        1: {
+            "name": "Text",
+            "texthl": "LanguageClientText",
+        },
+        2: {
+            "name": "Read",
+            "texthl": "LanguageClientRead",
+        },
+        3: {
+            "name": "Write",
+            "texthl": "LanguageClientWrite",
+        },
+    }
+
 ==============================================================================
 3. Commands                                           *LanguageClientCommands*
 
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -424,6 +424,18 @@ Signature: LanguageClient#textDocument_rangeFormatting(...)
 
 Format selected lines.
 
+*LanguageClient#textDocument_documentHighlight()*
+*LanguageClient_textDocument_documentHighlight()*
+Signature: LanguageClient#textDocument_documentHighlight(...)
+
+Highlight usages of the symbol under the cursor.
+
+*LanguageClient#clearDocumentHighlight()*
+*LanguageClient_clearDocumentHighlight()*
+Signature: LanguageClient#clearDocumentHighlight()
+
+Clear the symbol usages highlighting.
+
 *LanguageClient#workspace_symbol()*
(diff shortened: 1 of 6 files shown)


autozimu:
Great work!

Haven't looked into too much detail yet. Will check back some time today.

doc/LanguageClient.txt
```diff
@@ -281,6 +281,26 @@ neovim >= 0.3.0.
...
+
+Default: >
+    {
+        1: {
+            "name": "Text",
+            "texthl": "LanguageClientText",
+        },
+        2: {
+            "name": "Read",
+            "texthl": "LanguageClientRead",
```
autozimu: I believe with almost all theme there, there won't be such highlight groups defined. Can you change this highlight these groups to make it more useful?

src/languageclient.rs
```diff
@@ -256,6 +264,80 @@ impl State {
...
+        let document_highlight: Option<Vec<DocumentHighlight>> =
+            serde_json::from_value(result.clone())?;
+        if let Some(document_highlight) = document_highlight {
+            let highlights = document_highlight
+                .into_iter()
+                .map(|DocumentHighlight { range, kind }| Highlight {
+                    line: range.start.line,
+                    character_start: range.start.character,
+                    character_end: range.end.character,
+                    group: self.documentHighlightDisplay[&kind
```
autozimu: Please change from index to `get()`, of which the former is hard crash in case index is not found.

autozimu:
> An alternative would be to not clear the highlights automatically and instead add a function to clear them.

I'd agree with this approach, which is kind of more aligned with expectation from user.

For vim support, it can be achieved with `addmatchpos`, which is less powerful comparing with `nvim_buf_add_highlight`, but can be left out in this PR.

YaLTeR:
Addressed the comments, also added missing docs and added Document Highlight into the context menu.

autozimu:
👍  Could you please rebase the changes?

YaLTeR:
Done.

autozimu:
Thanks.

Issue #403, opened 2018-04-26
Should LanguageClient use softtabstop instead of tabstop?


hrantzsch:
If I understand `:help tabstop` correctly, it is not very much recommended to set `tabstop` to any value other than 8. But LangaugeClient-neovim uses `tabstop` to determine the value of `tabSize` in a [Document Formatting Request](https://microsoft.github.io/language-server-protocol/specification#document-formatting-request--leftwards_arrow_with_hook).

As suggested by the `tabstop` help, I configured `softtabstop` and `shiftwidth` in my vim config. Consequently, the indent created by autoformatting does not match my configured manual indent.

Should LanguageClient-neovim be sending the value of `softtabstop` with the command? If not, what is the recommended configuration?

autozimu:
Seems make sense.

From reading the doc, it feels like `shiftwidth` is the actual value used for indenting,
```
						*'shiftwidth'* *'sw'*
'shiftwidth' 'sw'	number	(default 8)
			local to buffer
	Number of spaces to use for each step of (auto)indent.  Used for
	|'cindent'|, |>>|, |<<|, etc.
	When zero the 'ts' value will be used.  Use the |shiftwidth()|
	function to get the effective shiftwidth value.
```

PR #566, opened 2018-08-14
Make rangeFormatting tab behavior same as regular

YaLTeR:
I noticed this when trying out range formatting for the first time. It's the same change as was requested for normal formatting in #403.


autozimu:
Thanks.

Issue #560, opened 2018-08-11
Make `LanguageClient#serverStatus()` clearer


liuchengxu:
## Is your feature request related to a problem? Please describe.

Currently, `LanguageClient#serverStatus()` returns 1 when the server is busy, 0 when the server is idle. 

If I have this mapping in my .vimrc:

```vim
    nnoremap <silent> gd :call LanguageClient#textDocument_definition()<CR>
```

When editing some file which doesn't have a language server support, e.g., vim, `gd` just tells me `Language server is not running for: vim`. `LanguageClient#serverStatus()` is 0 in this case, which is a little bit confusing(no language server running at all or it's still actually doing its job?).

What I want is when there is no language server running, fall back to the default behavior of `gd`.

```vim
" Suppose -1 indicates no language server defined/running
if LanguageClient#serverStatus() == -1
  call searchdecl(expand('<cword>'))
else
  call LanguageClient#textDocument_definition()
endif
```

## Describe the solution you'd like

The solution is to provide an option to tell people whether there is a language server exists for the current buffer, e.g., let `LanguageClient#serverStatus()` return -1 when the language server is not avaliable.

## Describe alternatives you've considered

For now, I can only tell based on `g:LanguageClient_serverCommands`. If the current file type is defined in `g:LanguageClient_serverCommands`, then I am able to assume that it has a language server support. It would be better `LanguageClient#serverStatus()` is clearer.

## Additional context
Suppose `LanguageClient#serverStatus()` returns -1 when a certain file doesn't have a language server running in the background, I could have a more robust `GoToDefinition()`:

```vim
function! GotoDefinition() abort
  if LanguageClient#serverStatus() == -1
    call searchdecl(expand('<cword>'))
  elseif LanguageClient#serverStatus() == 1
    echo 'Language Server is busy now, please try again later.'
  else
    call LanguageClient#textDocument_definition()
  endif
endfunction
```

autozimu:
The function you're looking for is `LanguageClient#isAlive`. You could use
```vim
:echomsg LanguageClient_runSync('LanguageClient#isAlive')
```
to check if server is running synchronously.

Or even better, I'd suggest use async way for your use case,
```vim
function GotoDefinitionHandler(output) abort
    if has_key(a:output, 'error')
       # fallback strategy
    endif
endfunction

function! GotoDefinition() abort
    call LanguageClient#textDocument_definition({'handle': v:true}, 'GotoDefinitionHandler')
endfunction

```

[...]

PR #568, opened 2018-08-14
Don't assign a Funcref to a local variable

liuchengxu:
Also a little bit refactor.

Ref: https://github.com/autozimu/LanguageClient-neovim/issues/560#issuecomment-412755693

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -273,7 +273,8 @@ function! s:HandleMessage(job, lines, event) abort
                 let l:Handle = get(s:handlers, l:id)
                 unlet s:handlers[l:id]
                 let l:type = type(l:Handle)
-                if l:type == s:TYPE.funcref
+                if l:type == s:TYPE.funcref ||
+                      \ (l:type == s:TYPE.string && exists('*' . l:Handle))
                     call call(l:Handle, [l:message])
                 elseif l:type == s:TYPE.list
                     call add(l:Handle, l:message)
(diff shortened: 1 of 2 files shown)


autoload/LanguageClient.vim
```diff
@@ -263,15 +270,14 @@ function! s:HandleMessage(job, lines, event) abort
                 endtry
             elseif has_key(l:message, 'result') || has_key(l:message, 'error')
                 let l:id = get(l:message, 'id')
-                " Function name needs to begin with uppercase letter.
                 let l:Handle = get(s:handlers, l:id)
                 unlet s:handlers[l:id]
-                if type(l:Handle) == type(function('tr')) ||
-                            \ (type(l:Handle) == type('') && exists('*' . l:Handle))
```
autozimu, line 270: What is the motivation to remove this line?

It supports using function name, i.e., string type, as handler.

liuchengxu: It's not typical to accept a string as a Funcref argument. This branch should only handle the Funcref case and we should not encourage people to pass a string as Funcref, they are definitely different and misleading. In a word,  always use `function('Goto')` instead of `'Goto'`.

autozimu: I don't think that's convincing.

It may not be typical and less commonly used. But it's been supported at this stage. Better to keep it this way.

liuchengxu: Done.

autozimu:
Thanks.

PR #571, opened 2018-08-16
Add complete funcref support for callback

liuchengxu:
This PR is actually the continuation of #568, I didn't realize that there are many other `l:callback` at that time :(, which should not work either when it comes to Funcref type. Since I seldom use these funcations and they are not tested very well, please review more carefully :). Let me know if something is still missed.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -451,30 +451,31 @@ function! LanguageClient#Notify(method, params) abort
 endfunction
 
 function! LanguageClient#textDocument_hover(...) abort
-    let l:callback = get(a:000, 1, v:null)
+    let l:Callback = get(a:000, 1, v:null)
     let l:params = {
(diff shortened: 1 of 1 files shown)


autoload/LanguageClient.vim
```diff
@@ -451,16 +451,15 @@ function! LanguageClient#Notify(method, params) abort
 endfunction
 
 function! LanguageClient#textDocument_hover(...) abort
-    let l:callback = get(a:000, 1, v:null)
```
autozimu, line 454: Instead of inline all the operations of getting the function, can you instead change the variable to upper case? i.e., `l:Callback` should work.

liuchengxu: Good catch. Have reverted all the `l:callback` to `l:Callback`.

autozimu:
Thanks.

PR #581, opened 2018-08-24
Update rust dependencies

dimbleby:
`cargo update`, and fix resulting breakages.

No special reason, just seemed like a nice thing to do.

Probably ought to provide a value for `hierarchical_document_symbol_support`?  But I figure that `None` is the safe thing for this pull request to do, and I leave it to you to say whether the client actually does support that - whatever it is.


autozimu:
Thanks.

PR #584, opened 2018-08-28
Fix debugInfo function formatting

kirillbobyrev:
Thanks for implementing the feature!

I believe, the function name should be enclosed in `*` so that the formatting style matches the rest of documentation.


autozimu:
Thanks.

PR #597, opened 2018-09-05
Add pre-built binary support for armv7* and armv8* targets

### LINKED ISSUE #596: Add (Linux) ARM targets to pre-built binaries
Anyway we could add ARM targets to the pre-built binary list; specifically `armv8`: `aarch64-unknown-linux-gnu`... Is it as simple as adding the target triplet to [line 17](https://github.com/autozimu/LanguageClient-neovim/blob/29d8ef8841b849437ea2e47b60f2581b2fd81f5e/.travis.yml#L17) of the Travis build config? I would be happy to open a PR...

It is pretty annoying to have to compile on my Chromebook (which takes about 15ish minutes) every time I want to update the plugin via a plugin manager...

PS very cool that you are migrating to Rust! Have you measured the performance improvement over Python yet?

mwilliammyers:
Close #596

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -45,6 +45,8 @@ arch=$(uname -sm)
 case "${arch}" in
     "Linux x86_64") download $name-$version-x86_64-unknown-linux-musl ;;
     "Linux i686") download $name-$version-i686-unknown-linux-musl ;;
+    "Linux aarch64") download $name-$version-aarch64-unknown-linux-gnu ;;
+    "Linux armv7"*) download $name-$version-armv7-unknown-musleabihf ;;
     "Darwin x86_64") download $name-$version-x86_64-apple-darwin ;;
     "FreeBSD amd64") download $name-$version-x86_64-unknown-freebsd ;;
     *) echo "No pre-built binary available for ${arch}."; try_build ;;
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
 matrix:
   include:
     # Linux
-    - env: TARGETS=i686-unknown-linux-musl:x86_64-unknown-linux-musl:i686-pc-windows-gnu:x86_64-pc-windows-gnu:x86_64-unknown-freebsd
+    - env: TARGETS=i686-unknown-linux-musl:x86_64-unknown-linux-musl:armv7-unknown-linux-musleabihf:aarch64-unknown-linux-gnu:i686-pc-windows-gnu:x86_64-pc-windows-gnu:x86_64-unknown-freebsd
 
     # OSX
     - env: TARGETS=x86_64-apple-darwin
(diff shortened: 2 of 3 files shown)


install.sh
```diff
@@ -45,6 +45,8 @@ arch=$(uname -sm)
 case "${arch}" in
     "Linux x86_64") download $name-$version-x86_64-unknown-linux-musl ;;
     "Linux i686") download $name-$version-i686-unknown-linux-musl ;;
+    "Linux aarch64") download $name-$version-aarch64-unknown-linux-gnu ;;
+    "Linux armv7"*) download $name-$version-armv7-unknown-musleabihf ;;
```
autozimu, line 49: Is this `*` intentional?

mwilliammyers: Yeah, from Googling it, it looks like `uname -m` reports `armv7l` and maybe `armv7`, `armv7hf`,  `armv7h`, and `armv7m` . It looks like it has to do with what type of floating-point precision the kernel is reporting from the CPU. Since rust only targets one armv7 arch with two libc ABIs (musl and glibc), I am pretty sure this will work, but I could change it to be `armv7l` knowing that it will work with the possibility of missing other armv7 architectures...

It was really tricky to piece together what `uname -m` would report vs what ARM lists on their wiki etc.. https://en.wikipedia.org/wiki/Uname says `armv7l` for sure...

mwilliammyers: Also, as far as I understand it, `rustc` automatically statically links musl libcl for triples with musl but creates a dynamically linked one for glibc (gnu, triples)... is that correct? i.e. there will not be a run-time dependence on musl libc which is why you had x86 use the musl triple?

autozimu: OK, just keep this `arv7*` for now until someone complains.

Yes, the difference is runtime dependencies on libc or not. Please update this to use musl.

mwilliammyers: I wanted to use musl but it looks like `cross` doesn't support it yet for `aarch64`... rust-embedded/cross#168

mwilliammyers: If I have time I can submit a PR to cross to add support for it...

autozimu: Ahh, you're right. I will merge this in.

autozimu: Good luck for the CR to cross!

Really wish rust cross compiling would become better.

mwilliammyers: Ok, thanks! I will submit a PR to update it to `musl` once cross adds support for it.

Also, just so you are aware, there are some OSes that will report `Linux aarch64` and `Linux armv7l` but do not have the same ABI that the `aarch64-unknown-linux-gnu` and `armv7-unknown-linux-musleabihf` triplets require. Android is one that I know of for sure (it requires `aarch64-linux-android`), but there are probably others - which is why I updated INSTALL.md to note that...

mwilliammyers: I thought about adding a check based on `uname -o` for `GNU/Linux` but I figured there could potentially be exotic distros that report something else for `uname -o` that run on the Linux kernel... Also, it ruins the nice one-line switch statement :)

autozimu: Good to know these. Thanks.

PR #606, opened 2018-09-14
Store documentHighlight buffer

YaLTeR:
I was hitting an issue that switching to a different window (or otherwise to a different buffer) was leaving document highlights hanging indefinitely. This was due to the clear notification being processed after the window was switched, resulting in `nvim_buf_clear_highlight` targeting the wrong buffer and doing nothing.

This MR makes documentHighlight store both the highlight source and the buffer ID so that `nvim_buf_clear_highlight` always targets the correct buffer.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -331,19 +331,26 @@ impl State {
                     })
                 }).collect::<Result<Vec<_>>>()?;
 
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks.

PR #615, opened 2018-09-22
make the preivew window close automatically

framlog:
a.t.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,15 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "dirs"
+version = "1.0.3"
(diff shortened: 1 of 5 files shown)


afnanenayet:
Why did you switch from using `env` for finding the home directory to the `dirs` crate?

framlog:
@afnanenayet because that `std::env::home_dir` is deprecated.

dimbleby:
Should at the very least separate the `home_dir` thing into its own commit I'd think.  Anyway, see #620 for discussion of why this may be unwanted.

See also #645 for discussion about closing the preview window.

afnanenayet:
Ah I see, I think I agree that switching to the new crate should be a separate PR/issue,  given the concerns that were raised in #620

framlog:
@dimbleby Thanks for your notice. I don't think this PR is needed after reading the author's view.

Issue #614, opened 2018-09-22
Add option to force build from source on install


samuelallan72:
## Is your feature request related to a problem? Please describe.

The `install.sh` script tries to download the pre-built release before attempting to build from source. This doesn't allow me to prefer building from source and avoiding pre-built binaries.

## Describe the solution you'd like

An argument for `install.sh` that disables attempting to download pre-built binaries. Something like `install.sh --no-binary` would be nice.

## Describe alternatives you've considered

I _could_ just tell the plugin manager to run `make release` instead of `install.sh`, but this doesn't allow for possible vim-specific installation instructions in the future.

autozimu:
> possible vim-specific installation instructions

What are those specific here?

The `install.sh` script is indented to be as portable as possible, (macOS, Linux, BSD). I'd like to keep it simple.

samuelallan72:
> What are those specific here?

There aren't right now, but if I modify the vim plugin manager to run `make release` to build from source (which seems to be the only thing needed from the `install.sh`), then if something changes in `install.sh` in the future, it won't be run when I next update.

By all means, keep the install script as simple as possible, but please allow a cli argument that forces it to build from source instead of pull pre-built files. :)

(I'm happy to work on a PR if you'd be interested in accepting such a change.)

autozimu:
That's not convincing enough. Right now there is no vim specific stuffin the `install.sh` script and I'm not indent to add any in the future.

Thanks for trying to help. But it's always quite a few back-and-forths to keep shell script portable for any small change.

samuelallan72:
I never said there was vim specific stuff in the install script. I merely said I didn't want the install script to try pulling the pre-built binaries first. This is primarily for security reasons since I would prefer to trust the source code instead of uploaded binaries.

autozimu:
In that case, back to the original question, please use `make release`.

The `install.sh` will install the prebuilt binary only and any post-install steps will go into the `release` make target.

PR #623, opened 2018-10-02
document building from source on install

samuelallan72:
ref issue #614


autozimu:
This has already been documented at https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#3-install-this-plugin 

Thanks anyway.

samuelallan72:
Oh so it is.

PR #626, opened 2018-10-07
Don't include full debugging info in the release build

tbodt:
Including debug information makes the binaries over 60 MB, which is unacceptable. Binaries without full debug info are around 5 MB.


autozimu:
Thanks.

PR #648, opened 2018-11-04
Retain history in quickfix and location lists, where appropriate

dimbleby:
If I do a search that populates the quickfix list, and then do a second search - then the quickfix list is replaced.

But sometimes it would be convenient to have a history of my search results (per `chistory`, `colder`, `cnewer`).

On the other hand when publishing diagnostics I think it *does* make sense to replace the existing list.

So the first commit in this pull request adds a parameter to the `setqflist()` and `setloclist()` methods, arranging that only when we publish diagnostics do we set the "r"-for-replace flag.

But then I noticed that although I now have the expected history, all of my lists are called `setqflist` - which isn't very helpful!  So the second commit puts a title on the quickfix or location list.  (There's clearly scope to bikeshed the exact text, and also perhaps it would be nice to distinguish the different types of search in future).

The third commit is just a `cargo fmt`.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -464,12 +464,13 @@ impl State {
                     }).collect::<Vec<_>>()
             }).collect();
 
+        let title = "[LC]: diagnostics";
         match self.diagnosticsList {
             DiagnosticsList::Quickfix => {
-                self.setqflist(&qflist, "r")?;
+                self.setqflist(&qflist, "r", title)?;
             }
             DiagnosticsList::Location => {
-                self.setloclist(&qflist, "r")?;
+                self.setloclist(&qflist, "r", title)?;
             }
             DiagnosticsList::Disabled => {}
         }
@@ -631,7 +632,7 @@ impl State {
         Ok(())
     }
 
-    fn display_locations(&mut self, locations: &[Location]) -> Result<()> {
+    fn display_locations(&mut self, locations: &[Location], title: &str) -> Result<()> {
         let location_to_quickfix_entry =
             |state: &mut Self, loc: &Location| -> Result<QuickfixEntry> {
                 let filename = loc.uri.filepath()?.to_string_lossy().into_owned();
@@ -680,7 +681,7 @@ impl State {
                     .map(|loc| location_to_quickfix_entry(self, loc))
                     .collect();
                 let list = list?;
-                self.setqflist(&list, " ")?;
+                self.setqflist(&list, " ", title)?;
                 self.echo("Quickfix list updated.")?;
             }
             SelectionUI::LocationList => {
@@ -689,7 +690,7 @@ impl State {
                     .map(|loc| location_to_quickfix_entry(self, loc))
                     .collect();
                 let list = list?;
-                self.setloclist(&list, " ")?;
+                self.setloclist(&list, " ", title)?;
                 self.echo("Location list updated.")?;
             }
         }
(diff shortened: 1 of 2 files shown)


dimbleby:
Oh, yuck, so the reason for the failing test is that the API used to be:

```
setqflist({list} [, {action}[, {title}]])
```

and that's what the aging neovim 0.1.7 in CI expects.  But that more recently it is as I have used it:

```
setqflist({list} [, {action}[, {what}]])
```

which is tedious to say the least.

Any views on whether there's a sensible way to handle this?  Or should I just give up on the title?

dimbleby:
looks like the old variant still works (it just isn't documented any more)

autozimu:
Sorry for the late response. I looked into the changes briefly but got distracted.

Did you test this changes on vim? I believe for vim, the API syntax is
```
setqflist({list} [, {action}[, {what}]])
```
which might also explain why neovim updated the API.

src/vim.rs
```diff
@@ -276,15 +276,19 @@ impl State {
         Ok(())
     }
 
-    pub fn setqflist(&mut self, list: &[QuickfixEntry]) -> Result<()> {
-        if self.call::<_, u8>(None, "setqflist", json!([list, "r"]))? != 0 {
+    pub fn setqflist(&mut self, list: &[QuickfixEntry], replace: bool, title: &str) -> Result<()> {
```
autozimu: I'd suggest making the parameter a bit more extensible, i.e., replace `replace: bool` to `action: &str`.

src/languageclient.rs
```diff
@@ -2698,7 +2719,8 @@ impl State {
         }
 
         let locations: Vec<Location> = serde_json::from_value(result.clone())?;
-        self.display_locations(&locations)?;
+        let title = format!("[LC]: rustdoc for {}", word);
```
autozimu: The message should be `... implementations ...`, not `... rustdoc ...`.

autozimu:
Got this error with vim 8.1
```
Error: ErrorMessage { msg: "\'Vim(let):E715: Dictionary required\'" }
```

dimbleby:
Using the dictionary argument to `setqflist()` was my first draft - that's what recent versions of neovim expect and document.  But that failed CI, because CI uses an old neovim that expects the string and does not support the dictionary.

I now believe that all versions of vim (that are sufficiently modern to support this plugin at all) expect the dictionary.

Specifically: neovim seems to accept a dictionary as at [0.2.0](https://github.com/neovim/neovim/blob/v0.2.0/runtime/doc/eval.txt#L2232) - and the relevant vim patch is [7.4.2200](https://github.com/vim/vim/commit/d823fa910cca43fec3c31c030ee908a14c272640).

So:
* could check "is_nvim" and, if so, use the string
* could just use the dictionary, and drop support for old versions of neovim.  This would require updating the neovim in CI
* other?

What do you think?

src/languageclient.rs
dimbleby: done

src/vim.rs
dimbleby: done

autozimu:
I can update neovim in CI.

dimbleby:
The awkward two-step approach is necessary for compatibility:

- if you provide the dictionary then the list is ignored
- only in more recent Vim and Neovim are you allowed to provide "items" in the dictionary
  - (especially, the Neovim 0.2.2 now in CI doesn't support this)

So that's why I've first set the list and then added the title, rather than doing it all in one step.

autozimu:
Thanks.

PR #653, opened 2018-11-08
Cargo update

dimbleby:
Just updating things while I'm in the area.

Only slightly interesting change is that `rpc::Notification` now has just `Params` rather than `Option<Params>` - which makes sense, since there's a `Params::None` value that can be used to express no-params already.


autozimu:
Thanks.

Issue #620, opened 2018-09-25
warning: use of deprecated item 'std::env::home_dir'


herrbischoff:
## Describe the bug

When compiling with `rustc 1.29.1 (b801ae664 2018-09-20)`, the following warning is output:

```
cargo build --release                                                                   
   Compiling languageclient v0.1.120 (file:///home/herrbischoff/.config/nvim/plugged/Lan
guageClient-neovim)                                                                     
warning: use of deprecated item 'std::env::home_dir': This function's behavior is unexpe
cted and probably not what you want. Consider using the home_dir function from https://c
rates.io/crates/dirs instead.                                                           
    --> src/languageclient.rs:2771:28                                                   
     |                                                                                  
2771 |                 let home = env::home_dir().ok_or_else(|| err_msg("Failed to get home dir"))?;                                                                            
     |                            ^^^^^^^^^^^^^                                         
     |                                                                                  
     = note: #[warn(deprecated)] on by default
```

## Environment
- neovim/vim version (`nvim --version` or `vim --version`): `NVIM v0.3.1`
- This plugin version (`git rev-parse --short HEAD`): `288f1be`
- This plugin's binary version (`bin/languageclient --version`): `languageclient 0.1.120 288f1beef601f8287be480ba144654ca91f86a8f`
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue. Refer to an example [here][min-vimrc.vim]): n/a
- Language server link and version: n/a

[min-vimrc.vim]: https://github.com/autozimu/LanguageClient-neovim/blob/next/min-vimrc.vim

## Expected behavior
Probably follow the suggestion given in the warning in exchanging the deprecated function for a supported one, to avoid breaking changes in the future.

autozimu:
Thanks.

I've noticed this warning. The reason I haven't make the change is that the replacement lib doesn't support BSD, which is one of the target of this project.

autozimu:
Or put it more accurately, the lib doesn't support cross-compiling to BSD.

PR #654, opened 2018-11-09
fix compile warnning, auto open quickfix or location-list when selection ui is not fzf

ghost:
fix compile warnning, auto open quickfix or location-list when selection ui is not fzf


dimbleby:
I've thought about this myself, but perhaps not everyone wants this behaviour.  So maybe a better approach would be to emit a custom event on completing a request (`doautocmd User ...`), which users could match on if they cared to do so?

See #620 for reasons not to fix the compiler warning.

cairijun:
Perhaps an option for this? I didn't notice this pr until I made this: https://github.com/richardtsai/LanguageClient-neovim/commit/eee73c6a2414afe360a92448eb7ee86b78f25658

autozimu:
The compile warning has been resolved.

Please add an option to allow customize this behavior.

autozimu (changes requested):
Please check previous comment for adding an option for this behaviour change.

Cosson2017:
Already added！

ghost:
@autozimu 
Already added！

PR #666, opened 2018-11-28
Improve symbol display for denite

ficoos:
* Aligns to actual columns instead of using \t
* Shows symbol's kind
* Searches the symbol name instead of the whole display line
* Highlights the results

documentSymbol
workspaceSymbol

### DIFF
--- a/rplugin/python3/denite/source/documentSymbol.py
+++ b/rplugin/python3/denite/source/documentSymbol.py
@@ -1,19 +1,16 @@
 from typing import List, Dict
+from os.path import dirname
+import sys
 
 from .base import Base
 
+sys.path.insert(0, dirname(dirname(__file__)))
 
-def convert_to_candidate(symbol: Dict, bufname: str) -> Dict:
-    name = symbol["name"]
-    start = symbol["location"]["range"]["start"]
-    line = start["line"] + 1
-    character = start["character"] + 1
-    return {
-        "word": "{}:{}:\t{}".format(line, character, name),
-        "action__path": bufname,
-        "action__line": line,
-        "action__col": character,
-    }
+from common import (  # isort:skip  # noqa: I100
+    convert_symbols_to_candidates,
+    SYMBOL_CANDIDATE_HIGHLIGHT_SYNTAX,
+    highlight_setup,
+)
 
 
 class Source(Base):
@@ -23,9 +20,11 @@ class Source(Base):
         self.name = 'documentSymbol'
         self.kind = 'file'
 
+    def highlight(self):
+        highlight_setup(self, SYMBOL_CANDIDATE_HIGHLIGHT_SYNTAX)
+
     def gather_candidates(self, context: Dict) -> List[Dict]:
         bufname = self.vim.current.buffer.name
         result = self.vim.funcs.LanguageClient_runSync(
             'LanguageClient_textDocument_documentSymbol', {}) or []
-        return [convert_to_candidate(symbol, bufname) for
-                symbol in result]
+        return convert_symbols_to_candidates(result, bufname)
--- a/rplugin/python3/denite/source/workspaceSymbol.py
+++ b/rplugin/python3/denite/source/workspaceSymbol.py
@@ -1,12 +1,15 @@
-from urllib import request, parse
 from os import path
-from typing import List, Dict
+import sys
 
 from .base import Base
 
+sys.path.insert(0, path.dirname(path.dirname(__file__)))
 
-def uri_to_path(uri: str) -> str:
-    return request.url2pathname(parse.urlparse(uri).path)
+from common import (  # isort:skip  # noqa: I100
+    convert_symbols_to_candidates,
+    SYMBOL_CANDIDATE_HIGHLIGHT_SYNTAX,
+    highlight_setup,
+)
 
 
 class Source(Base):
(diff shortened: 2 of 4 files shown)


whitebyte:
Mmm, neat stuff!

rplugin/python3/denite/source/common.py
```diff
@@ -0,0 +1,131 @@
+from typing import List, Dict
```
autozimu: If this file is common utility rather than a source, maybe it's better to move into a different dir?

rplugin/python3/denite/source/common.py
```diff
@@ -0,0 +1,131 @@
...
+    },
+]
+
+
+def uri_to_path(uri: str) -> str:
+    return request.url2pathname(parse.urlparse(uri).path)
+
+
+def highlight_setup(source, syntax):
+    def mangle_name(n):
```
autozimu: Missing argument and return types. Also, can you find a more self-explanatory name than `n` or else please add a doc.

rplugin/python3/denite/source/common.py
```diff
@@ -0,0 +1,131 @@
...
+        max_kind_len = max(max_kind_len, len(kinds[-1]))
+        candidates.append({
+            "word": name,
+            "action__path": filepath,
+            "action__line": line,
+            "action__col": character,
+        })
+
+    for c, p, k in zip(candidates, paths, kinds):
+        c["abbr"] = "{:<{}} [{:^{}}] {}".format(
```
autozimu: Please avoid using single letter variable names.

rplugin/python3/denite/source/documentSymbol.py
```diff
@@ -1,19 +1,16 @@
 from typing import List, Dict
+from os.path import dirname
+import sys
 
 from .base import Base
 
+sys.path.insert(0, dirname(__file__))
```
autozimu: With subsequent invoke, would the sys.path keep increasing?

rplugin/python3/denite/source/workspaceSymbol.py
```diff
@@ -17,28 +20,33 @@ def __init__(self, vim):
...
-            'LanguageClient#workspace_symbol', '', {}) or []
-        return self.convert_to_candidates(result)
+            'LanguageClient#workspace_symbol', prefix, {}) or []
+        if current_buffer != bufnr:
+            self.vim.command("tabclose")
+
+        candidates = convert_symbols_to_candidates(
+            result,
+            pwd=self.vim.funcs.getcwd())
+        if len(result) < self.max_candidates:
```
autozimu: Where is this max_candidates come from?

rplugin/python3/denite/source/workspaceSymbol.py
```diff
@@ -17,28 +20,33 @@ def __init__(self, vim):
...
+        context['is_interactive'] = True
+        prefix = context['input']
+        bufnr = context['bufnr']
+
+        cached_prefix = context.get('__cached_prefix')
+        if cached_prefix is not None and prefix.startswith(cached_prefix):
+            return context['__cached_candidates']
+
+        current_buffer = self.vim.current.buffer.number
+        if current_buffer != bufnr:
```
autozimu, line 40: Just curious, under which case would this becoming true?

rplugin/python3/denite/source/documentSymbol.py
ficoos: No, this only happens on import and imports happen only once

rplugin/python3/denite/source/workspaceSymbol.py
ficoos: from [Base](https://github.com/Shougo/denite.nvim/blob/cb1daf74c51b1670ee3914c95f85ba3852525e17/rplugin/python3/denite/source/base.py#L19)

rplugin/python3/denite/source/workspaceSymbol.py
ficoos: LanguageClient does everything in the context of the current buffer. When interactively filtering results the current buffer is the denite buffer not the source buffer. This means that LanguageClient keeps returning empty results because there is no server defined for denite buffers (obviously). This is a trick to open a new tab with the original buffer and then run the command in that buffer.

The proper way to fix is to have LanguageClient have APIs to run in the context of a buffer or a specific server, but it would have been much more complicated than this hack and I (and others) need this to work now as most projects have more than a 1000 symbols.

I will document this in the code and open an issue requesting these APIs.

(edit) not->no

rplugin/python3/denite/source/common.py
ficoos: Where would you have me put it?

autozimu: Ideally,
```
.
├── common.py
└── source
    ├── codeAction.py
    ├── contextMenu.py
    ├── documentSymbol.py
    ├── references.py
    └── workspaceSymbol.py
```

rplugin/python3/denite/source/workspaceSymbol.py
autozimu: I see.

Although, from the documentation and implementation, it seems like max_candidates are only applied to candidates filtering,
- https://github.com/Shougo/denite.nvim/blob/a59c3e2f50452e0ddd1988e921624f6cfb47807d/doc/denite.txt#L1957
- https://github.com/Shougo/denite.nvim/blob/0caf719e590bd53c165cdb3f8fdb752f0c336bba/rplugin/python3/denite/denite.py#L161

ficoos: It appears you are correct. It just so happens that ccls just chose the same number as the limit of it's results.

Looking at the protocol I don't see a way for the server to notify the client that the result set is not complete.

I will change it to always ask the server no matter how many results.

ficoos:
@autozimu, Is there anything else you want me to fix?

autozimu:
Sorry for the late response. Been busy these days.

Will take another look shortly.

autozimu:
Thanks!

PR #669, opened 2018-12-03
Fix completion where text_edit is supplied

ficoos:
The code mistakenly tries to compare the line instead of the character.
Even the comment says we assume the lines are equal.


autozimu:
Thanks.

PR #670, opened 2018-12-03
Use denite in contextMenu

ficoos:
When using denite for the context menu, the actions should be opened
in denite.

### DIFF
--- a/rplugin/python3/denite/source/contextMenu.py
+++ b/rplugin/python3/denite/source/contextMenu.py
@@ -2,6 +2,14 @@ from typing import List, Dict
 from .base import Base
 
 
+DeniteOverrides = {
+    'Code Action': 'codeAction',
+    'Document Symbol': 'documentSymbol',
+    'Workspace Symbol': 'workspaceSymbol',
+    'References': 'references',
+}
+
+
 class Source(Base):
     def __init__(self, vim):
         super().__init__(vim)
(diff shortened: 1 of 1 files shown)


rplugin/python3/denite/source/contextMenu.py
```diff
@@ -10,10 +18,17 @@ def __init__(self, vim):
         self.kind = 'command'
 
     def convert_to_candidate(self, item):
+        if item in DeniteOverrides:
+            cmd = 'call denite#start({})'.format(
+                repr([{"name": DeniteOverrides[item],
```
autozimu: Is this `repr` necessary here?

ficoos: For lists `str()` calls `repr()` on the underlying objects so the effect is identical. But the *intent* is to use `repr()`. If it wasn't a list, maybe a string for example, `repr()` would have been required.

This documents the fact that we *want* the `repr()` of the object and not the `str()` and ignores the fact that for lists coincidentally `str() == repr()`.

autozimu: My point here is, `repr` is not necessary, nor `str`, as the goal is formatting one string.

How about this?
```python
cmd = 'call denite#start(["name": {}, "args": []])'.format(DeniteOverrides[item])
```

ficoos: Oh, I didn't get that this is what you meant. I guess you are right.

autozimu:
Thanks.

PR #671, opened 2018-12-03
Make LanguageClient_runSync reentrant

ficoos:
By changing `LanguageClient_runSync_outputs` from being source local to
being function local we gurantee that if the function is called from
multiple threads the function will still work.

This also documents that this variable *is* function local and should
not be referenced outside this function.


autozimu:
Thanks.

PR #672, opened 2018-12-05
fix: relative path in install.ps1

Priddles:
The relative path to `bin\` does not refer to the intended directory when
the working directory is not the same as the root of this project.

This causes `install.ps1` to fail when using a plugin manager.


autozimu:
Thanks

PR #678, opened 2018-12-19
Improved sign and highlight updates when multiple buffers are displayed

TheUnknownCylon:
Hi,

This PR improves sign and highlight updates when multiple buffers are displayed at the same time.

Old behavior, with 2 buffers shown to the user. When a change in buffer 2 (Book.java) creates an error for another buffer (1, PrintABook.java), the quickfix list is immediately updated. However no sign and highlights are added to the line in the other buffer. Signs and highlights are only added when moving to the other buffer, and making a change there (triggering LC). See the GIF below for a demo:
demo_old

New behavior: with this PR not only the quickfix list is updated immediately. Also the signs and highlights are updated when a change triggers an error in another buffer. See the GIF below for a demo of the new behavior, and notice that no movement in the other buffer is required to get the sign and highlight to be added or removed: 
demo

Some remarks:
* These are my first steps in Rust, and the first steps in creating a PR for a VIM plugin. Feel free to give any feedback.
* In case I missed this implementation on another PR/branch, please feel free to reject the PR :)
* I did split up the `languageClient_handleCursorMoved`, moving sign display code, and highlight display code to their own methods for improved readability, and better reuse-ability.
* Signs and highlights are set for the entire buffer, not just the visible region. Signs and highlights are set for all _open_ buffers. Not sure how this affects the performance for large files.
* This also implies that the signs column in always visible if there are marks _anywhere_ in the buffer (imho this is fine, this is what other plugins (sometimes?) do as well)
* Following the development instructions, tests did pass :) Apparently there are some formatting issues, but some seem to be unrelated to my changes... :( 

I think it's a useful improvement, please let me know what you think. Do you agree with the added functionality? Do you agree with the implementation itself?

Kind regards,
TheUC

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -190,6 +190,18 @@ function! s:AddHighlights(source, highlights) abort
     for hl in a:highlights
         call nvim_buf_add_highlight(0, a:source, hl.group, hl.line, hl.character_start, hl.character_end)
     endfor
+  endfunction
+
+function! s:ClearHighlightsByFile(filename, source) abort
+    let bufid = bufnr(a:filename)
+    call nvim_buf_clear_highlight(bufid, a:source, 0, -1)
+endfunction
+
+function! s:AddHighlightsByFile(filename, source, highlights) abort
+    let bufid = bufnr(a:filename)
+    for hl in a:highlights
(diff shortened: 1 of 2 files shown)


TheUnknownCylon:
Closed PR because there seems to be no interest for it.

PR #682, opened 2018-12-23
Use proper filetype when showing hover info in preview

cairijun:
Currently the preview window is opened with `filetype=markdown` in all cases, which might cause mis-highlighting for non-markdown content. LSP has explicit statements about where markdown are used, including the deprecated `MarkedString` which has no `language` specified, and the new `MarkupContent` when `kind = "markdown"`. This PR addresses these specifications on a best effort. An exception is that when multiple `MarkedString`s are returned, `markdown` is always used as vim cannot have multiple filetypes in a single buffer.

### DIFF
--- a/src/languageclient.rs
+++ b/src/languageclient.rs
@@ -857,17 +857,21 @@ impl State {
         Ok(())
     }
 
-    fn preview<S>(&mut self, lines: &[S]) -> Fallible<()>
+    fn preview<D>(&mut self, to_display: &D) -> Fallible<()>
     where
-        S: AsRef<str> + Serialize,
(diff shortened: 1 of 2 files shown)


autozimu:
Can you run `cargo fmt` to fix this CI build failure?

cairijun:
Done. BTW, on my env `cargo fmt` gives completely different results from those on CI. Not sure if the version of `rustfmt` matters. Mine is `rustfmt 1.0.0-stable (4a4e5081 2018-11-29)`.

autozimu:
Thanks.

Issue #586, opened 2018-08-29
Buffer not created properly when jump into definition for a second time (buffer not showing in the tab line)


dylan-chong:
- Did you upgrade to latest plugin version? Yes
- Did you upgrade to/compile latest binary? Run shell command 
  `bin/languageclient --version` to get its version number. languageclient 0.1.114 542203b6aa435b93a21e123b2177b18cd1a1a8ed
- (Neovim users only) Did you check output of `:checkhealth LanguageClient`? Yes everything is fine
- Did you check [troubleshooting]?

[troubleshooting]: https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#6-troubleshooting

## Describe the bu

Vim buffer may not properly be created when jumping to the definition of the file. This causes the buffer not be shown in the airline tab line

## Environment
- neovim/vim version (`nvim --version` or `vim --version`): NVIM v0.3.1
- This plugin version (`git rev-parse --short HEAD`): 891ee2c
- This plugin's binary version (`bin/languageclient --version`): languageclient 0.1.114 542203b6aa435b93a21e123b2177b18cd1a1a8ed
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue. Refer to an example [here][min-vimrc.vim]):
```
call plug#begin('~/.vim/plugged')

Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }
Plug 'vim-airline/vim-airline'

call plug#end()

augroup filetype_rust
    autocmd!
    autocmd BufReadPost *.rs setlocal filetype=rust
augroup END

" Always draw sign column. Prevent buffer moving when adding/deleting sign.
set signcolumn=yes

let g:LanguageClient_serverCommands = {
    \ 'rust': ['rustup', 'run', 'stable', 'rls'],
    \ 'typescript': ['typescript-language-server', '--stdio'],
    \ }
let $RUST_BACKTRACE = 1
let g:LanguageClient_loggingLevel = 'INFO'
let g:LanguageClient_loggingFile = stdpath('data') . '/LanguageClient.log'
let g:LanguageClient_serverStderr = stdpath('data') . '/LanguageServer.log'

au BufRead,BufNewFile *.ts set filetype=typescript

let g:airline#extensions#tabline#enabled = 1

```
[...]

PR #683, opened 2018-12-26
Ensure buflisted after jumping to a location

cairijun:
If a buffer was "deleted" via `:bdelete`, it was not actually completed deleted, and still can be queried via `bufnr`. `s:Edit` currently uses `buffer` to jump to an existing buffer, but `buffer` does not add the buffer to the buflist, which is the root cause of #586. `set buflisted` should ensure the buffer we just jumped into presents in the buflist normally.


autozimu:
Thanks

dylan-chong:
Thanks @richardtsai !

PR #686, opened 2018-12-28
Create factory method for HighlightDefinition

### LINKED ISSUE #685: LanguageClient requires Python3.7
The current version of the LanguageClient requires Python3.7 as of dfdf91743e0246063e16526ef3ead4342a15c886. The reason that `defaults` parameter of `namedtuple` is used, which was only added in Python3.7. This requirement is nowhere stated and rather harsh in my opinion as Python3.7 is still very new.
I am personally using intelpython, where the newest version is only 3.6.5.

Please consider dropping the `defaults` parameter and make it compatible with Python3.5, or at least state the requirement of Python3.7.

ficoos:
The `default` argument for `collections.namedtuple` is only available
from python 3.7. Since we don't want to have this the only thing
breaking support with earlier python versions we use an explicit factory
method to get the same result.

Fixes #685

### DIFF
--- a/rplugin/python3/denite/common.py
+++ b/rplugin/python3/denite/common.py
@@ -13,24 +13,35 @@ from lsp.protocol import SymbolKind  # isort:skip  # noqa: I100
 
 MAX_FNAME_LEN = 30
 
-HighlightDefinition = namedtuple("HighlightDefinition", (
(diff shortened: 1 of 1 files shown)


autozimu:
Thanks

PR #687, opened 2018-12-30
Remove redundant `[LC]` prefix

dimbleby:
echomsg_ellipsis always adds the prefix, so we were getting two of them

Edit: for reference, [here](https://github.com/autozimu/LanguageClient-neovim/blob/f9a9123582abfd0f0433e7f8813daf7e615a18b6/autoload/LanguageClient.vim#L44) is where the prefix is added.


autozimu:
Thanks

PR #688, opened 2018-12-30
Update dependencies

dimbleby:
In particular, updating language client types picks up version 3.14.0 of
the specification and we're forced to handle a couple of additions:
- Add support for signature label offsets
- Explicitly don't add support for location links

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ name = "aho-corasick"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


dimbleby:
I think you must have an old version of rustfmt, presumably on your docker image?  I have

    rustfmt 1.0.0-stable (4a4e508 2018-11-29)

I believe that rustfmt is supposed to be stable as at 1.0.0, so it probably makes more sense for you to upgrade on the docker image, rather than for me to downgrade and reformat?

autozimu:
rustfmt is indeed outdated in the docker image.

I will update it when I get a stable internet connection.

dimbleby:
(rebased, re-updated, and re-formatted)

autozimu:
Thanks

PR #702, opened 2019-01-24
Fix: types::State can not be serialized as json

cairijun:
c101096f0289cd932e84ee3c12a2624b19ffe499 intruduced a `clients` `HashMap` in `types::State`, whose key is `LanguageId`, which makes `types::State` can not be serialized as json, and it breaks some vim functions like `LanguageClient#getState()`. One example is that Airline can not display diagnostic info from LanguageClient, and keeps poping up errors instead. This PR skip serializing this field as it seems not being used in vim script currently.


autozimu:
Thanks

PR #706, opened 2019-01-28
added sync version LanguageClient#textDocument_formatting

dbarrosop:
Not sure if there is a better way but this is useful to autoformat the code on save with:

```
au BufWritePre *.py,*.go :call LanguageClient#textDocument_formatting_sync()
```

Otherwise, the async nature of the call will make nvim to save the buffer before the code has actually been formatted.


autozimu:
Thanks!

PR #710, opened 2019-01-29
Allow user override of 'has snippet support'

### LINKED ISSUE #709: snippet support broken at 657ad707
Prior to 657ad707, I had snippet support working fine with ultisnips and ncm2-ultisnips.

But after 657ad707, the language client does not recognise that ultisnips supports snippets.  So it declares that I don't have snippet support, and no server will play along.

To be fair I did have to do a (small) amount of customisation to get this working, so the language client is not completely wrong when it says that snippets are not supported.  But I *have* made it work, and now it doesn't!

Sadly I can't override the language client's opinion: setting `g:LanguageClient_hasSnippetSupport = 1` doesn't turn on snippet support - though it sounds as though it ought to.  As it stands, that variable would be better re-named `g:LanguageClient_doNotDisableSnippetSupport`.

Can we please make this work again?  Maybe with a new setting so that I can force snippet support on, maybe by changing the meaning of the existing setting, maybe something else altogether...?

dimbleby:
Fixes #709

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -103,8 +103,8 @@ function! s:Debug(message) abort
 endfunction
 
 function! s:hasSnippetSupport() abort
-    if get(g:, 'LanguageClient_hasSnippetSupport', 1) !=# 1
-        return 0
+    if exists('g:LanguageClient_hasSnippetSupport')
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks!

PR #714, opened 2019-01-30
Fix: registerHandlers does not register callback function correctly

cairijun:
`registerHandlers` accepts callbacks as function name strings, and serializing the names again makes them being registered as `"function_name"`, which makes the callbacks do not get called correctly.


autozimu:
Thanks!

PR #716, opened 2019-01-30
fix comparison with v:null

### LINKED ISSUE #715: Cannot disable virtual text
The comparison with `v:null` in the recent commit introducing virtual text is faulty, so that it's not possible to disable this feature.  I'll submit a pull request fixing it shortly.

dimbleby:
Fixes #715


dimbleby:
I fixed a couple more examples in the codebase while I'm at it.

To be explicit: the problem is that `0` is considered equal to `v:null` if you compare with `!=` etc.

autozimu:
Thanks!

PR #734, opened 2019-02-16
Show the most severe diagnostic in the gutter (not least)

TysonAndre:
(i.e. if there is both an Error and Information for a given line,
show the sign for Error on the left hand side)

Previously, this was showing the least severe - it was adding the integer value for the sort (in Sign's id property), not subtracting it

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -569,8 +569,20 @@ impl LanguageClient {
             })
             .collect();
 
-        // There might be multiple diagnostics for one line. Show only highest severity.
-        signs.sort_unstable();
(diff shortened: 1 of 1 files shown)


TysonAndre:
I fixed the formatting. I don't have any experience with rust programming, so if there's a clearer way to do this, let me know. I wasn't sure if changing the id would break the order of diagnostics, but I suspected it would, so I wrote a separate lambda to sort this

autozimu:
Actually I'm working a sign display rehaul at dev branch, which would make this change obsolete.

Thanks for contribution and sorry for this inconvenience.

PR #749, opened 2019-03-03
Fix build warning

h-michael:


src/language_server_protocol.rs
```diff
@@ -2033,7 +2033,7 @@ impl LanguageClient {
                                     notify::RecursiveMode::NonRecursive
                                 };
                                 watcher.watch(
-                                    w.glob_pattern.trim_right_matches("**"),
+                                    w.glob_pattern.trim_start_matches("**"),
```
dimbleby: `trim_end_matches()`, surely?

TysonAndre:
the issue causing the build on this PR to fail should be fixed:

```patch
cargo fmt -- --check
Diff in /root/project/src/language_server_protocol.rs at line 2032:
                                 } else {
                                     notify::RecursiveMode::NonRecursive
                                 };
-                                watcher.watch(
-                                    w.glob_pattern.trim_end_matches("**"),
-                                    recursive_mode,
-                                )?;
+                                watcher
+                                    .watch(w.glob_pattern.trim_end_matches("**"), recursive_mode)?;
```

src/vim.rs
```diff
@@ -66,10 +57,6 @@ impl Vim {
         self.rpcclient.notify("cursor", json!([lnum, col]))
     }
 
-    pub fn setline(&self, lnum: u64, text: &[String]) -> Fallible<()> {
```
autozimu, line 69: This is unused right now, but I think it is very likely to be used in the short near future.

If the warning is disturbing you, please add `[allow(unused)]` to turn if off instead.

autozimu:
I merged in a huge refactor days ago and rendered this PR unmergeable.

Please update it you still want it merged in. Sorry for the inconvenience and thanks for contribution.

h-michael:
We don't need these changes now :)

PR #751, opened 2019-03-03
Bumpup jsonrpc-core from v9.0.0 to v10.1.0

h-michael:
[jsonrpc(not only jsonrpc-core) diff: v9.0.0...v10.1.0](https://github.com/paritytech/jsonrpc/compare/v9.0.0...v10.1.0)


autozimu:
Thanks.

PR #755, opened 2019-03-05
Implement window/showMessageRequest

fsouza:
I probably did a lot of things wrong, feedback is appreciated :)

I was able to test this with Scala Metals and everything seems to work fine (see https://github.com/fsouza/vimfiles/commit/914bb5d994e81e93634d215fdb93000f317c1443).

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1841,9 +1841,9 @@ impl LanguageClient {
         let mut v = Value::Null;
         let result: Option<i64> = self.vim()?.rpcclient.call("s:inputlist", options)?;
         if let Some(answer) = result {
-            if answer > 0 && answer < (n_actions - 1) as i64 {
-                let raw_actions: Vec<Value> = try_get("actions", &params)?.unwrap_or_default();
-                v = raw_actions[(answer - 1) as usize].clone();
+            let raw_actions: Vec<Value> = try_get("actions", &params)?.unwrap_or_default();
+            if let Some(action) = raw_actions.get((answer - 1) as usize) {
+                v = action.clone();
             }
         }
         info!("End {}", lsp::request::ShowMessageRequest::METHOD);
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1824,6 +1824,32 @@ impl LanguageClient {
         Ok(())
     }
 
+    pub fn window_showMessageRequest(&self, params: &Value) -> Fallible<Value> {
+        info!("Begin {}", lsp::request::ShowMessageRequest::METHOD);
+        let msg_params: ShowMessageRequestParams = params.clone().to_lsp()?;
+        let msg_actions = msg_params.actions.unwrap_or_default();
+        let n_actions = msg_actions.len();
+        let actions: Vec<Value> = msg_actions
+            .into_iter()
+            .enumerate()
+            .map(|(i, item)| Value::String(format!("{}) {}", i + 1, item.title)))
+            .collect();
+        let mut options: Vec<Value> = Vec::with_capacity(actions.len() + 1);
+        options.push(Value::String(msg_params.message));
+        options.extend(actions);
+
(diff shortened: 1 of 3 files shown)


src/language_server_protocol.rs
```diff
@@ -1824,6 +1824,32 @@ impl LanguageClient {
...
+            .map(|(i, item)| Value::String(format!("{}) {}", i + 1, item.title)))
+            .collect();
+        let mut options: Vec<Value> = Vec::with_capacity(actions.len() + 1);
+        options.push(Value::String(msg_params.message));
+        options.extend(actions);
+
+        let mut v = Value::Null;
+        let result: Option<i64> = self.vim()?.rpcclient.call("s:inputlist", options)?;
+        if let Some(answer) = result {
+            if answer > 0 && answer < (n_actions - 1) as i64 {
```
autozimu: The second portion should be something like `<= n_actions`, right?

Say there are 2 actions, `2` should be a legitimate answer.

src/language_server_protocol.rs
```diff
@@ -1824,6 +1824,32 @@ impl LanguageClient {
...
+                let raw_actions: Vec<Value> = try_get("actions", &params)?.unwrap_or_default();
```
autozimu: Parse action is not necessary since there is already deserialized instance before hand. Around L1833, you can use `iter()`  instead of `into_iter()` so you can continue use the instance here.

src/language_server_protocol.rs
```diff
@@ -1824,6 +1824,32 @@ impl LanguageClient {
...
+                v = raw_actions[(answer - 1) as usize].clone();
```
autozimu: Please try not indexing from an array. If the index is out of range, it crashes the whole program. use `get()` instead.

autozimu:
Thanks for contributing. Please see my comments inline.

src/language_server_protocol.rs
fsouza: Yeah I thought the struct in lsp-type wouldn't contain everything that's required, but I just checked the spec and it does. Will fix the code.

src/language_server_protocol.rs
fsouza: We're doing a bounds check two lines before 😬 I'll update the code though.

src/language_server_protocol.rs
fsouza: The mapping between actions and parameters to ``inputlist`` isn't 1 to 1, they're actually shifted to accommodate the prompt. ``params.message`` is actually the index 0, so when the user types 1, they're actually selecting 0.

See screenshot:

image

We're instructing the user to type from 1 to 3, but the user could actually type 0 and get an invalid option. Vim also supports mouse click, so that's why we can't simply say that the second option is index 0.

src/language_server_protocol.rs
fsouza: Question: what's the easiest way to convert ``MessageActionItem`` into ``Value``? 🤔

src/language_server_protocol.rs
autozimu: Yes, I get that part.

What I meant is, say in this screenshot, if I select or click 3, it should still be an acceptable answer, right? But in your code, `3 < (3 - 1)`, which is returning false.

src/language_server_protocol.rs
autozimu: `serde_json::to_value`

src/language_server_protocol.rs
fsouza: Oh shoot you're right I got this wrong, my bad.

autozimu:
Thanks!

fsouza:
@autozimu hey I didn't get a chance to address all the feedback, sorry about that. I can look into them later today or tomorrow.

Thanks for all the feedback!

autozimu:
I will create a commit to clean it up.

fsouza:
Oh cool, thank you and sorry about that: :)

PR #766, opened 2019-03-12
Use :normal! instead of :normal

rhysd:
Hi,

In my `init.vim`, `m` mapping is overridden and I found that `LanguageClient#textDocument_definition` was broken on my local. After some investigation, I noticed that `:normal` was used in this plugin. Plugin should use `:normal!` instead of `:normal`.


autozimu:
Thanks!

PR #767, opened 2019-03-12
Reimplement hover window with floating window for Neovim 0.4.0 or later

rhysd:
Hi,

I have experimentally reimplemented a hover window for `LanguageClient#textDocument_hover()` with [Neovim's new floating window feature](https://github.com/neovim/neovim/pull/6619).

Here is a screencast with Neovim 0.4.0-dev and macOS 10.12:

test

This PR is still work in progress. I have not added any tests yet and I think there are some discussion points as follows:

- There is no way to move cursor into float window so user cannot scroll the contents. Some mappings might be needed.
- There is no way to close float window by user (No command like `:pclose` is not provided at this point). I added functionality to close the floating window automatically when cursor is moved. But I'm not sure this behavior makes all users happy
- Some people may prefer preview window. Using floating window might be opt-in or opt-out
- `CursorLine` is used for highlighting the floating window, but I'm not sure it is a good choice to modify background color slightly for every colorscheme. The highlight color might be customizable

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -263,7 +263,8 @@ function! s:GetVar(...) abort
 endfunction
 
 function! s:ShouldUseFloatWindow() abort
-    return s:FLOAT_WINDOW_AVAILABLE && get(g:, 'LanguageClient_useFloatingHover', 1)
+    let use = s:GetVar('LanguageClient_useFloatingHover')
+    return s:FLOAT_WINDOW_AVAILABLE && (use || use is v:null)
 endfunction
 
 function! s:CloseFloatingHoverOnCursorMove(win_id, opened) abort
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -262,7 +262,7 @@ function! s:GetVar(...) abort
     endif
 endfunction
 
-function! s:CloseFloatingHover(bufname, opened) abort
+function! s:CloseFloatingHoverAfterCursorMove(bufname, opened) abort
     if getpos('.') == a:opened
         " Just after opening floating window, CursorMoved event is run.
         " To avoid closing floating window immediately, check the cursor
@@ -270,14 +270,25 @@ function! s:CloseFloatingHover(bufname, opened) abort
         return
     endif
     autocmd! plugin-LC-neovim-close-hover
-    let bufnr = bufnr(a:bufname)
-    if bufnr == -1
-        return
-    endif
-    let winnr = bufwinnr(bufnr)
+    let winnr = bufwinnr(bufnr(a:bufname))
     if winnr == -1
         return
     endif
+    execute winnr . 'wincmd c'
+endfunction
+
+function! s:CloseFloatingHoverAfterEnterAnotherWin(win_id) abort
+    let winnr = win_id2win(a:win_id)
+    if winnr == 0
+        " Float window was already closed
+        autocmd! plugin-LC-neovim-close-hover
+        return
+    endif
+    if winnr == winnr()
+        " Cursor is moving into floating window. Do not close it
+        return
+    endif
+    autocmd! plugin-LC-neovim-close-hover
     execute winnr . 'wincmd c'
 endfunction
 
@@ -293,9 +304,9 @@ function! s:OpenHoverPreview(bufname, lines, filetype) abort
 
         " Unlike preview window, :pclose does not close window. Instead, close
         " hover window automatically when cursor is moved.
-        let params = printf('"%s", %s', a:bufname, string(pos))
+        let call_after_move = printf('<SID>CloseFloatingHoverAfterCursorMove("%s", %s)', a:bufname, string(pos))
         augroup plugin-LC-neovim-close-hover
-            execute 'autocmd CursorMoved,CursorMovedI,InsertEnter <buffer> call <SID>CloseFloatingHover(' . params . ')'
+            execute 'autocmd CursorMoved,CursorMovedI,InsertEnter <buffer> call ' . call_after_move
         augroup END
 
         " Calculate width and height and give margin to lines
@@ -333,7 +344,7 @@ function! s:OpenHoverPreview(bufname, lines, filetype) abort
             let col = 1
         endif
 
-        call nvim_open_win(bufnr('%'), v:true, width, height, {
+        let float_win_id = nvim_open_win(bufnr('%'), v:true, width, height, {
         \   'relative': 'cursor',
         \   'anchor': vert . hor,
         \   'row': row,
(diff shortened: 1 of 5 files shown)


autozimu:
Awesome!

theHamsta:
Nice work!

Currently, one should not switch buffers (e.g. <c-o>) when hover is visible. The float is only closed when moving the cursor in the original buffer.

rhysd:
I added test for this feature (though I'm not sure it's sufficient).

> There is no way to move cursor into float window so user cannot scroll the contents. Some mappings might be needed.

Regarding to this, I feel providing a mapping to move cursor into floating window would be enough. User can scroll the window and yank contents by moving cursor to the window.

> Currently, one should not switch buffers (e.g. ) when hover is visible. The float is only closed when moving the cursor in the original buffer.

Thank you for the point. I think `WinLeave` should trigger closing float window as well. But it is also run when moving cursor into floating window. I think `WinEnter` would be available instead. Let me try.

rhysd:
I noticed `WinEnter` is not sufficient since it cannot close floating window when user switches current buffer to another buffer within the same window. I changed it to `BufEnter` and added one more test.

autozimu:
Is this ready for review?

rhysd:
Let me add a mapping to move cursor into a floating window. Tomorrow is holiday in Japan. So I think I can have some progress on this branch tomorrow (I'm sorry for the delay).

autozimu:
Not at all. Please take your pace.

rhysd:
I finished the work.

Instead of adding a new mapping, I slightly changed the behavior of `LanguageClient#textDocument_hover()`. When opening a hover and calling it again before moving cursor, it moves the cursor into the floating window. Then user can scroll content and yank some text. `:close` or `:quit` in the window simply close the hover window.

test

I also added

- `g:LanguageClient_useFloatingHover`
- Documentation
- More test for scenario moving the cursor into floating window

Now I believe this branch is ready for review :+1:

lorenzo:
I think having another function we can call to force the hover to be in a separate panel would be ideal. While most of the times I would use this floating window, there are also times that I want to leave the docs for a function open as I keep coding. Having both possibilities would make this the ideal feature

rhysd:
Sounds good. It may be more useful than moving cursor into floating window but I'm not sure. Anyway, I'll try to implement the function to close floating window and reopen the content in preview window again.

rhysd:
I implemented `LanguageClient#reopenHoverInSeparateWindow()`. Could you check the behavior?

lorenzo:
Why is it `reopenIn` and not just `openIn`?

rhysd:
> Why is it reopenIn and not just openIn?

Since it is only effective when floating window is already open. If `openIn` is more natural, I'll revert the commit. I'm not native English user. So the naming may be not proper. If my English usage is not proper, please let me know. It's very helpful for me (in terms of learning English).

JeanMertz:
I think what @lorenzo was getting at is that it perhaps makes more sense to have two functions:

* `openHoverInFloatingWIndow` – this opens the "hover" definition in the new floating window
* `openHoverInRegularWindow` – this opens the "hover" definition in a regular window

But I'm not sure it makes sense to have this. There's already `LanguageClient_useFloatingHover`, so you could have your own function that first sets `LanguageClient_useFloatingHover=0` and then calls `LanguageClient#textDocument_hover()`, to achieve the exact same result, I think?

Unless of course, I'm mistaken, and that's not what @lorenzo is talking about.

I like the `reopenHoverInSeparateWindow` though, it allows to quickly swap to a separate window if you're in a hover, and realise you want to keep this documentation open while you continue working.

lorenzo:
The idea is not to have your own function, as few people actually write vimL. I don’t, and have been a bum user for years.

I did mean to have 2 separate function to open the panel in different places. For example, you could keep one panel open while using a floating window for something else as you write a long function.

Moving a floating window to a panel does make sense to me as well, I can see the use case.

rhysd:
Thank you for further explanation. I got the point.

Two options come in my mind for this.

### Add argument to hover function

I feel creating multiple functions is a bit too much. How about adding an argument for specifying window type?

- `LanguageClient#textDocument_hover()`: Auto (use floating window as much as possible
- `LanguageClient#textDocument_hover('preview')`: Open in preview window
- `LanguageClient#textDocument_hover('float')`: Open in floating window

Separate two mappings can be defined such as `nnoremap K :<C-u>call LanguageClient#textDocument_hover()<CR>` and `nnoremap gK :<C-u>call LanguageClient#textDocument_hover('preview')`. It may be useful to add another type such as `'vsplit'` (opens new normal window by `:vsplit`).

- Pros:
  - Simple and stateless
- Cons:
  - User needs to define two mappings

### Add special window-local mapping

Defines buffer local (floating window local) mapping `p` to reopen the window in separate window.

User opens floating window with `K` and moves cursor by entering `K` again. User enter `p` to open the document in separate window. It may be more useful to provide another local mapping such as `q` for closing the window.

- Pros:
  - It does not require new mapping
- Cons:
  - Operation is complicated. User needs to enter the floating window to reopen it in separate window

---

I prefer second one because I would almost always use floating window and opening it in separate window would be comparably rare case.

In terms of implementation, either (or both) is easy to implement.

asymmetric:
I’m not sure the use-case of permanently opening the hover/tooltip is necessary enough to block this otherwise very useful PR. I’ve never found this behavior in other editors and personally never expected it.

Can I suggest we merge this and iterate over the “permanent hover” behavior in a separate PR?

autozimu:
Agree with @asymmetric. Let's have the support first. Then we can cover the special use case.

autoload/LanguageClient.vim
```diff
@@ -261,6 +262,154 @@ function! s:GetVar(...) abort
     endif
 endfunction
 
+function! s:ShouldUseFloatWindow() abort
+    return s:FLOAT_WINDOW_AVAILABLE && get(g:, 'LanguageClient_useFloatingHover', 1)
```
autozimu: Can you change it to make use of `s:GetVar`? which would support buffer local variable as well.

doc/LanguageClient.txt
```diff
@@ -63,6 +63,12 @@ accessed by regular quickfix/location list operations.
 To use the language server with Vim's formatting operator |gq|, set 'formatexpr': >
     set formatexpr=LanguageClient#textDocument_rangeFormatting_sync()
 <
+If you're using Neovim 0.4.0 or later, |LanguageClient#textDocument_hover()|
```
autozimu: I'd suggest move this section to function documentation as it is function specific.

rhysd:
@asymmetric @autozimu 

OK, let me remove the part for now.

rhysd:
@autozimu I addressed your review comments at 79e36ba and a785c80.

autozimu:
Thanks!

YaLTeR:
I just tried this and it works amazingly! This is so much better than what was possible previously, thanks a lot.

One thing I noticed: when the hover window is opened, Ctrl-E and Ctrl-Y don't move it with the buffer so it kind of becomes detached from the cursor. Not sure if this would be easy (or in scope) to fix though.

rhysd:
Yeah, it's a bit difficult case. Currently floating window is closed automatically on `CursorMoved`, `BufEnter` and `InsertEnter`. But `<C-e>` and `<C-y>` is none of the cases. It just scrolls screen and position of cursor is maintained. I don't know the way to hook `<C-e>` and `<C-y>` actually..

autoload/LanguageClient.vim
Zquinn10: Yes

Issue #765, opened 2019-03-12
Regression: Signs are displayed with diagnosticsSignsMax set to zero


mburakov:
## Description
Up to 0.1.140 it was possible to disable signs (and gutter) entirely by setting diagnosticsSignsMax to zero. This feature was introduced with bb733ef. Unfortunately, this does not work anymore since 0.1.141.

## To Reproduce
1. Set diagnosticsSignsMax to zero (`let g:LanguageClient_diagnosticsSignsMax = 0`)
2. Open any source file with warnings/errors
3. Observe gutter with corresponding signs

## Current behavior
Gutter and diagnostic signs are visible after setting diagnosticsSignsMax to zero.

## Expected behavior
Gutter and diagnostic signs must not be visible after setting diagnosticsSignsMax to zero.

## Additional context
I assume that since 0.1.141 **zero** value is mistakenly reinterpreted as **null** value. And displaying all the signs is a documented behavior for null value.

autozimu:
Resolved by https://github.com/autozimu/LanguageClient-neovim/pull/769

Issue #758, opened 2019-03-06
Regression: Signs do not disappear in gutter after line shifts.


mhuttner:
## Description
Warning signs in gutter do not get deleted properly after deleting or inserting a line. 
This was reintroduced in 0.1.141, 0.1.140 does not have this issue

## To Reproduce
- Open a valid python file, using python-language-server 
- Delete a line causing multiple syntax errors
- Undo the change, reinserting the line

## Current behavior
- Warning signs in the gutter will remain

## Expected behavior
- Resolved warnings are cleared

## Possible Cause
I think the problem was introduced by
https://github.com/autozimu/LanguageClient-neovim/blob/788e91931b9397cad047fed4c74811a62fdac1eb/src/sign.rs#L20-L23

Since the line number changes and causes the Equality check to fail.

AdaShoelace:
In addition to signs not disappearing when inserting or deleting a line, they also persist after correcting the error in question. This in turn forces me to call `signs unplace *` after correcting an error in order for the signs to reflect the correct state of my code.

kierun:
If anyone wants to see how it looks, this is it:

![screenshot](https://user-images.githubusercontent.com/3079575/54273496-e82cf600-457d-11e9-8593-a7180563050c.png)

Just wrote a simple script, making many errors along the way.

autozimu:
Should be resolved by https://github.com/autozimu/LanguageClient-neovim/pull/769

PR #769, opened 2019-03-13
fix diagnostic sign issues

cairijun:
This PR should fix two diagnostic signs related issues:
* #758 If multiple diagnostic messages were reported on a single line, multiple signs would be added but only one of them would be removed after the errors were resolved. I think we can just use the most severe one and drop the others  on one line.
* #765 Limitation on number of signs was dropped.

@mhuttner @PierreLeidbring @mburakov Mind helping verify if is this fix you issues?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,6 +160,11 @@ version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "either"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "failure"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -301,6 +306,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "either 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "itoa"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -334,6 +347,7 @@ dependencies = [
  "diff 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "itertools 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "jsonrpc-core 10.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazycell 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


AdaShoelace:
This seems to fix the issue with the "error crosses" not getting removed from the gutter. However the little "i" is still persisting even though there is no error, warning or other reason to.

mburakov:
Hi @richardtsai, thank you for taking a look. ~~I don't have any experience with rust, neither a prepared build environment. But I will be happy to try a prebuilt binary if you can share one.~~

It appeared to be easier than I expected. Your change fixes my issue. Thanks again!

cairijun:
> This seems to fix the issue with the "error crosses" not getting removed from the gutter. However the little "i" is still persisting even though there is no error, warning or other reason to.

@PierreLeidbring 
could you provide a reproducible example? code, language server & operations should be sufficient

cairijun:
I think I just found the problem. After lines were deleted, existing signs beyond the current viewport might jump in but be ignored by the viewport filter when calculating `signs_prev`. As the number of signs in `state.signs` is already limited by the viewport filter when adding new signs, I suppose we can just mark all the existing signs as obsolete without notable performance penality.

@PierreLeidbring Does the latest commit fix you issue?

AdaShoelace:
@richardtsai No it does not. Now it marks all my imports with the yellow triangle and the row on which the error used to occur still has a red cross even though the error is resolved.

cairijun:
I cannot reproduce the issues you mentioned, with either `clangd`, `pyls` or `rls`. Could you tell me which langserver you are using and a piece of example code? @PierreLeidbring

AdaShoelace:
To reproduce the error just create a new bin-project with `cargo new foo`.
Inside the project go down to the `println!("hello world!");`, press `shift + o`, write `String::` and wait for the error to occur. Remove the `String::` (which resolves the error). Now the `println!("Hello world!")`-line still have the red cross in the gutter.

From my vimrc:
```
			\ 'rust': ['rustup','run','nightly', 'rls'],
			\}

let g:LanguageClient_autoStart = 1
```

I tried using just `'rust': ['rls']` instead of `'rust': ['rustup','run','nightly', 'rls']`. That resolves the issue with the persisting signs but autocompletion stops working.

cairijun:
unable to reproduce:
https://dump.video/i/ZRXVLD.mp4

are you sure you are using the `languageclient` executable built from the commits in this PR?

AdaShoelace:
I think so at least. Installed it using:
```
Plug 'richardtsai/LanguageClient-neovim', {
			\ 'branch': 'fix_diag_signs',
			\ 'do': 'bash install.sh'
			\}

```
And updated with `:PlugUpdate` immediately after your first message today.

cairijun:
`install.sh` just downloads the official prebuilt executable
you have to `cargo build --release` followed by `mv target/release/languageclient bin/`

AdaShoelace:
My bad for not reading `install.sh`.
It seems to be working correctly with the latest fix.
Thanks!

cairijun:
@autozimu any thoughts? This seem to be blocking many users, expecially those of `clangd` or `pyls`.

autozimu:
Thanks!

sriharshachilakapati:
I had a temporary fix for this one till now at some point that I didn't notice this.

~~~vimscript
autocmd BufWritePre *.purs :sign unplace *
~~~

Thanks for this fix!!

jamesob:
Can you consider cutting a new release so that this fix is included in a regular `:PlugUpdate` workflow? If I hadn't dug through this issue thread and saw your note about manually building with `cargo`, I'd just uninstall your plugin permanently because the out-of-date gutter is so annoying.

autozimu:
This is included in latest release.

To be clear, you’re “free” to do whatever you like. I made this plugin
mostly for fun and published in hope it might be useful to some people,
which may or may not be including you. You’re definitely not obligated to
use it in any way.

jamesob:
Hm, well whether it's in the latest release or not, simply running `:PlugUpdate` (i.e. having the latest git clone of the `next` branch) didn't fix the problem; it was only resolved with the manual `cargo` stuff.

Sorry, didn't mean to come off so abrasively. Thank you for your work - it's very valuable and this kind of system improves my development experience quite a bit. I just wanted to emphasize that such a bug could scare off potential users quite easily.

autozimu:
Updating the repo itself won’t get the latest binary build. You had to run
the install script or let plugin manager to do the work.

jamesob:
Yeah, sure, but what I did was run `:PlugUpdate` which (if I understand correctly) should also run the `do:` post-update hook, which is `bash install.sh` - so something with that seems broken.

jamesob:
Yep, just confirmed that running `install.sh` causes a regression in this behavior (on 41c948e05532a92b27c9ef0509cea651592d6bb5).

Anrock:
`install.sh` only downloads latest binary. If there is no release cut then there is no binary for it and `install.sh` would not download anything.

jamesob:
> If there is no release cut then there is no binary for it and install.sh would not download anything.

I verified a new binary was downloaded when I ran `install.sh`. To be clear, the issue this PR fixes is still broken per the installation/update instructions in the README.

autozimu:
`install.sh` actually will always download the latest binary. I'm pretty sure this PR has been included in latest binary release, i.e., version 0.1.146.

@jamesob If it is "broken", can you file an issue with detailed reproduction steps? It's not clear what wasn't working from "broken".

PR #772, opened 2019-03-16
Don't print "Project root" message in Vim

JeanMertz:
There's already the option to print this message to a debug log if you want to, so I don't see the point of always printing `[LC] Project root: ...` in Vim on startup, even if you've configured the logger to only log `WARN` and `ERROR` messages.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -2751,7 +2751,6 @@ impl LanguageClient {
             .into()
         };
         let message = format!("Project root: {}", root);
-        self.vim()?.echomsg_ellipsis(&message)?;
         info!("{}", message);
         self.update(|state| {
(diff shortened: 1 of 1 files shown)


autozimu:
I'd say in most case, having this information echoed is handy when things not working as expected.

On the other hand, please add an option instead if you really don't want this behaviour.

JeanMertz:
> I'd say in most case, having this information echoed is handy when things not working as expected.

I agree. But there is already an option to see this information. The line below the one I removed, it calls the `info!` macro with the exact same message.

Isn’t that enough for debugging purposes?

autozimu:
It is enough, but then the user would need to find/specify where the log file is, and open the file to inspect.

autozimu:
Added option `LanguageClient_echoProjectRoot` to allow disable the echo.

PR #779, opened 2019-03-26
Use `target-cpu=native` when building release

alok:
This is if users end up building manually. It should give a free speedup since
Rust enables more CPU-specific features and does not affect compile time to my
knowledge.


alok:
That's weird. Git pulled a *lot* of commits on my local copy. Will close and re-submit.

PR #780, opened 2019-03-26
Use native CPU when building manually

alok:
This should give a free speedup in the vast
majority of cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ vint:
 	vint autoload plugin
 
 release:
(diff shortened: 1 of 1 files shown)


autozimu:
You can put this in your personal cargo config file, right?

If that's the case, then it's better to put it in the wiki instead of use it as default.

The same argument this isn't the default in cargo applies to here as well. With this setting, the compiled binaries might not be runnable from a different machine, even when the other machine is running exactly the same OS.

PR #790, opened 2019-04-14
Refactor code to use references where functions don't consume a value

afnanenayet:
I ran clippy on the codebase to look for some potential issues to fix. One of the suggestions was to use references for parameters that aren't consumed.


autozimu:
Thanks!

PR #794, opened 2019-04-18
Cross platform PowerShell install script

TylerLeonhardt:
This makes the PowerShell script work across Windows, macOS, and Linux for the cross-platform version of [PowerShell](https://github.com/PowerShell/PowerShell).

This is still backwards compatible with Windows PowerShell.


autozimu:
Thanks!

PR #796, opened 2019-04-21
Use self to avoid name repetition

afnanenayet:
I'm going through this project and trying to address all of the clippy warnings. This is a pretty basic one and easy to fix, instead of repeating the name of your struct, it is recommended that you use `Self` instead


autozimu:
Thanks!

PR #797, opened 2019-04-21
Remove unnecessary closures

afnanenayet:
For less complexity, you can remove unnecessary closures and simply pass in the function itself


autozimu:
Thanks!

PR #799, opened 2019-04-22
Add examples about how to configure for Ruby

guilleiguaran:
This adds an example about how to use LanguageClient-neovim with [Solargraph](https://github.com/castwide/solargraph) for Ruby.


autozimu:
Thanks!

sovetnik:
Can you add example with rvm too?

PR #801, opened 2019-04-25
Try to prepare sha256 and use it on install

spk:
Hi, I would like to prepare and use a `release-exe.sha256` file for each release and use it to check the checksum on the `install.sh`; I'm not sure about the result file and deploy process so maybe its a good template if its interesting for you ? cheers and thanks for this project!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,18 +11,21 @@ name=languageclient
 
 try_curl() {
     command -v curl > /dev/null && \
-        curl --fail --location "$1" --output bin/$name
+        curl --fail --location "$1" --output bin/$name && \
+        curl --fail --location "$1.sha256" --output bin/$name.sha256
 }
 
 try_wget() {
     command -v wget > /dev/null && \
-        wget --output-document=bin/$name "$1"
+        wget --output-document=bin/$name "$1" && \
+        wget --output-document=bin/$name.sha256 "$1.sha256"
 }
 
 download() {
     echo "Downloading bin/${name}..."
     url=https://github.com/autozimu/LanguageClient-neovim/releases/download/$version/${1}
     if (try_curl "$url" || try_wget "$url"); then
+        sha256sum -c "bin/$name.sha256" || (echo "bin/$name.sha256 file checksum does not match exiting." && exit 1)
         chmod a+x bin/$name
         return
     else


ci/before_deploy.sh
```diff
@@ -7,6 +7,9 @@ package() {
     fi
 
     cp -f target/$TARGET/release/$BIN_NAME bin/$BIN_NAME_TAG
+    cd bin || return 1
+    sha256sum "$BIN_NAME_TAG" | tee "$BIN_NAME_TAG.sha256"
```
autozimu: Thanks for the work!

Can you update this a little bit to
```
sha256sum "$BIN_NAME_TAG" | tee "bin/$BIN_NAME_TAG.sha256"
```
rather than changing dirs twice?

install.sh
```diff
@@ -11,18 +11,23 @@ name=languageclient
 
 try_curl() {
     command -v curl > /dev/null && \
-        curl --fail --location "$1" --output bin/$name
+        curl --fail --location "$1" --output bin/$name && \
+        curl --fail --location "$1.sha256" --output bin/$name.sha256
```
autozimu, line 15: Is it needed to cleanup previous sha256 before downloading? Like the current cleanup of bin file.

ci/before_deploy.sh
spk: Thanks sure done in 1509f42

install.sh
spk: With the `--output` from `curl` and `--output-document` from `wget` the file will be overwritten so I don't think its useful but if you think its safer let me know

install.sh
```diff
@@ -11,18 +11,21 @@ name=languageclient
...
-        wget --output-document=bin/$name "$1"
+        wget --output-document=bin/$name "$1" && \
+        wget --output-document=bin/$name.sha256 "$1"
 }
 
 download() {
     echo "Downloading bin/${name}..."
     url=https://github.com/autozimu/LanguageClient-neovim/releases/download/$version/${1}
     if (try_curl "$url" || try_wget "$url"); then
+        sha256sum -c "bin/$name.sha256" || return 1
```
autozimu: Sorry for the late update.

Can you please split this PR into two? One includes generating the SHA files and the other downloading and checking SHA. Since this is modifying the installation, erroneous script will resulting release unable to install.

Another issue, can you print an error message here before `return 1`?

Thanks again for your contribution.

spk: No worries, yep you're right i've squashed and apply the error message on daf3d2130ad98d71c524d435b4e8ac61896b6927 will create another PR for the generation (edit: https://github.com/autozimu/LanguageClient-neovim/pull/819), thanks for the feedback

autozimu: 👍

PR #802, opened 2019-04-25
Add cmd arg on install.sh script

spk:
This allow to download or compile as wanted
I've also changed the `sh` to `bash` because it have bashims in it let me know if not wanted
cheers


spk:
Not useful shebang fix extracted in https://github.com/autozimu/LanguageClient-neovim/pull/806

PR #803, opened 2019-04-25
Try to have idempotent install based on version checking

spk:
Hi, trying to have an idempotent install (maybe we can add an option to force install ?); cheers

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,7 +20,7 @@ try_wget() {
 }
 
 download() {
-    echo "Downloading bin/${name}..."
+    echo "Downloading bin/${name} ${version}..."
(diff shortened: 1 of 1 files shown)


autozimu:
Genuinely curious, what's the benefit of using this approach rather than the original? The original approach itself is idempotent, right?

spk:
> Genuinely curious, what's the benefit of using this approach rather than the original? The original approach itself is idempotent, right?

I'm currently using vim-plug and when updating with `PlugUpdate` it download every time a new version since its [deleted](https://github.com/spk/LanguageClient-neovim/blob/next/install.sh#L42) even if the binary is not updated or changed, maybe this is specific with vim-plug ?

autozimu:
Thanks!

PR #806, opened 2019-04-28
Use bash on shebang install.sh

spk:


install.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
autozimu, line 1: This is `sh` instead of `bash` is since BSDs don't have bash installed. Is there a strong reason this has to be changed?

spk: Oh I though it was bash specific since on the README there is `bash install.sh` sorry for the noise closing (cf https://github.com/autozimu/LanguageClient-neovim/commit/0ebcd3706e199a543c5cef08a0557c12f5c509a6)

PR #809, opened 2019-04-29
add new root marker for python projects: pyproject.toml

ticosax:
As defined per https://www.python.org/dev/peps/pep-0518/
`pyproject.toml` is a file that could be used to define a python project.

It is already used by some popular tools like [poetry](https://poetry.eustace.io/)
or [black](https://black.readthedocs.io/en/stable/)


autozimu:
Thanks!

PR #815, opened 2019-05-05
Add client-side support for "rust-analyzer.applySourceChange" 

JeanMertz:
The [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer) language server requires the clients to implement a custom source change operation.

This is similar to the already existing "java.apply.workspaceEdit" functionality in this language client.

This feature can be removed once rust-analyzer [supports `workspace/executeCommand`](https://github.com/rust-analyzer/rust-analyzer/issues/1232).

This also adds support for the `CreateFile` operation in `apply_WorkspaceEdit`, and makes sure `textDocument_codeAction` handles [all possible response values](https://github.com/gluon-lang/lsp-types/blob/d78071ae83367dda76c674e392a16268fe41e843/src/request.rs#L413).

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+# defaults are great :)
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1236,7 +1236,22 @@ impl LanguageClient {
             },
         )?;
 
-        let commands: Vec<Command> = serde_json::from_value(result.clone())?;
+        use lsp::request::CodeActionRequest;
+        let code_action_response: <CodeActionRequest as Request>::Result =
+            serde_json::from_value(result.clone())?;
+
+        let commands = match code_action_response {
+            None => return Ok(result),
+            Some(response) => match response {
+                CodeActionResponse::Commands(commands) => commands,
+                CodeActionResponse::Actions(actions) => {
+                    actions
+                        .into_iter()
+                        .filter_map(|action| action.command)
+                        .collect()
+                }
+            },
+        };
 
         let source: Vec<_> = commands
             .iter()
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -251,9 +251,11 @@ impl LanguageClient {
     }
 
     fn apply_WorkspaceEdit(&self, edit: &WorkspaceEdit) -> Fallible<()> {
+        use self::{DocumentChangeOperation::*, ResourceOp::*};
+
         debug!("Begin apply WorkspaceEdit: {:?}", edit);
-        let filename = self.vim()?.get_filename(&Value::Null)?;
-        let position = self.vim()?.get_position(&Value::Null)?;
+        let mut filename = self.vim()?.get_filename(&Value::Null)?;
+        let mut position = self.vim()?.get_position(&Value::Null)?;
 
         if let Some(ref changes) = edit.document_changes {
             match changes {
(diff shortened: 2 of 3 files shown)


src/language_server_protocol.rs
```diff
@@ -793,20 +804,61 @@ impl LanguageClient {
...
             }
-        } else {
-            bail!("Not implemented: {}", cmd.command);
+            "rust-analyzer.applySourceChange" => {
+                fn json_to_workspace_edit(
+                    json: Value,
+                ) -> Fallible<(WorkspaceEdit, Option<TextDocumentPositionParams>)> {
+                    let mut map: Map<String, Value> = from_value(json)?;
+                    let ws = from_value::<WorkspaceEdit>(
+                        map.remove("workspaceEdit").unwrap_or_default(),
```
autozimu, line 843: Have you considered creating a wrap type and using serde to convert Value to concrete type?

rustfmt.toml
```diff
@@ -0,0 +1 @@
+# defaults are great :)
```
autozimu, line 1: If defaults are great, is there a reason this file was added here?

JeanMertz: I added the reason in the commit message:

> By adding an (empty) `rustfmt.toml` file, you signal to developers and
code editors that `rustfmt` is required, and the default values are
used.

Many (large) projects have not yet adopted Rustfmt, so enabling it by default in your editor – if you work on many different open source projects – can result in unwanted commits/diffs when contributing to those projects.

So instead, what I do is only enable Rustfmt automatically if the editor detects a `rustfmt.toml` file in the root of the project. But aside from the potential for automation, it also signals to developers that this project uses Rustfmt, even though we use all the defaults.

There are [some other projects](https://github.com/search?q=filename%3Arustfmt.toml+defaults) that do the same, although admittedly not as much as I'd like.

JeanMertz: By the way, at first I also added a [`rust-toolchain` file](https://github.com/rust-lang/rustup.rs/blob/ed65847e14b86b0e6c34c4acfcc80fa386ce1f5a/README.md#the-toolchain-file) containing `stable` as its contents, as the nightly rustfmt and stable rustfmt behave differently with defaults. That would be another automated signals to tools to use the stable toolchain when working on this project.

Unfortunately, that broke the CI, so I removed it from the commit.

src/language_server_protocol.rs
JeanMertz: You mean doing the serialisation in one pass, instead of first serialising it to a `Map<String, Value>` and then separately serialising to `WorkspaceEdit` and `TextDocumentPositionParams`?

I have not. Do you maybe have some examples on how to do this? I'd gladly make the change 👍

autozimu: That's correct.

It should work similarly with other types defined in lsp-types and used in this repo. As an example, https://github.com/gluon-lang/lsp-types/blob/d78071ae83367dda76c674e392a16268fe41e843/src/lib.rs#L66-L71

The new type could be something like
```rust
struct WorkspaceEditWithCursorPosition {
    workspaceEdit: WorkspaceEdit,
    cursorPosition: TextDocumentPositionParams,
}
```

autozimu: Any updates on this?

JeanMertz: Sorry, not yet. It's on my list, I hope to get to it this weekend.

autozimu:
Thanks for contributing!

I cleaned up a bit and pushed those changes.

PR #830, opened 2019-06-02
edit install instructions deoplete plugin

robatipoor:
neovim completion not work for me we need set global variable deoplete#enable_at_startup = 1 in configure file init.vim


autozimu:
Thanks for contributing.

This makes me thinking that including deoplete setting is never a good idea in the first place, which itself would take various steps to make it working.

PR #842, opened 2019-07-02
Add missing space between severity, code, and message

amarshall:
Previously would render like “[ERROR]This is the error”, now renders like: “[ERROR] This is the error”.


autozimu:
Thanks!

PR #845, opened 2019-07-05
[#835] Set hover window as modifiable

### LINKED ISSUE #835: Preview windows does not continuously work
If the preview windows has already opened. Further `hover()` would display a empty preview windows. (Close preview windows first then call `hover()` works fine)

nvim --version
NVIM v0.3.7

- This plugin version (`git rev-parse --short HEAD`):

41c948e

- This plugin's binary version (`bin/languageclient --version`):

languageclient 0.1.146 ec4af74cc389f802c3c43a660c963049848a8a93

- Language server link and version:

any

YaLTeR:
If it's not set as modifiable and the preview window was already open as
nomodifiable (by the same function), then it fails setting the window text.

Fixes #835.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -370,7 +370,7 @@ function! s:OpenHoverPreview(bufname, lines, filetype) abort
         wincmd P
     endif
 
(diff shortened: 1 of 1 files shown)


autozimu:
Thanks!

PR #851, opened 2019-07-11
Expand tilde in linux and macos logger paths

martskins:
As it stands today, using tilde (`~`) in the `g:LanguageClient_loggingFile` path will break the server upon start. This PR adds a few lines to expand the tilde, for both linux and macos systems, using `shellexpand`, which was already included in the dependencies.


autozimu:
Thanks!

Issue #857, opened 2019-07-31
Completion includes types of variables/methods


vorner:
Maybe this is the same thing as #813? I'm not 100% sure.

- Did you upgrade to latest plugin version?

yes, commit `e4478a4789805c66ce24ae786677ae557cd411da`

- Did you upgrade to/compile latest binary? Run shell command
  `bin/languageclient --version` to get its version number.

```
$ ./bin/languageclient --version
languageclient 0.1.147
```
- (Neovim users only) Did you check output of `:checkhealth LanguageClient`?

```
  - OK: binary found: /home/vorner/.local/share/nvim/plugged/LanguageClient-neovim/bin/languageclient
  - OK: languageclient 0.1.147 
```

- Did you check [troubleshooting]?

Yes

## Describe the bug

I'm using ccls for the language client. The LanguageClient completes the code including types and parameters which don't actually belong to the code and need to be deleted.

It worked fine with 0.1.146 (only the relevant parts were added).

## Environment
- neovim/vim version (`nvim --version` or `vim --version`):

```
NVIM v0.3.7
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/x86_64-pc-linux-gnu-gcc -ggdb0 -O2 -pipe -march=native -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/var/tmp/portage/app-editors/neovim-0.3.7/work/neovim-0.3.7_build/config -I/var/tmp/portage/app-editors/neovim-0.3.7/work/neovim-0.3.7/src -I/usr/include -I/var/tmp/portage/app-editors/neovim-0.3.7/work/neovim-0.3.7_build/src/nvim/auto -I/var/tmp/portage/app-editors/neovim-0.3.7/work/neovim-0.3.7_build/include
Compiled by portage@leviathan

Features: +acl +iconv +jemalloc +tui 
See ":help feature-compile"

[...]

Issue #662, opened 2018-11-23
CompletionItem::textEdit support is incorrect


MaskRay:
## Describe the bug
When `CompletionItem.textEdit` is present but `CompletionItem.insertText` is absent, completion triggered by `triggerCharacter` does not display the completion item correctly.

```cpp
#include <bits/stdc++.h>
int main() { std::xplicit }
/*
atomic_fetch_and_explicit
*/
```

I've set `g:LanguageClient_completionPreferTextEdit` to 1

And IMHO `g:LanguageClient_completionPreferTextEdit` should just default to 1 if the server sends `CompletionItem.textEdit`.

## Environment
- neovim/vim version (`nvim --version` or `vim --version`): NVIM v0.3.2-649-g512e0caae
- This plugin version (`git rev-parse --short HEAD`): 13fc7e2 
- This plugin's binary version (`bin/languageclient --version`): languageclient 0.1.129
- Language server link and version: https://github.com/MaskRay/ccls/

## To Reproduce
1. Start vim
2. `let g:LanguageClient_completionPreferTextEdit = 1`
3. Insert boilerplate and `std::`
4. Select the first completion item
5. See error

## Current behavior
`newText` is somehow inserted from the line beginning though it does not overwrite existing characters.

## Expected behavior
`newText` should be inserted in the correct position, not from the line beginning.

JeanMertz:
I believe I am hitting the same issue, while using rust-analyser.

When I complete the selection at this point:

<img width="457" alt="screenshot 2019-02-14 at 08 23 12" src="https://user-images.githubusercontent.com/383250/52769851-d9602b80-3031-11e9-9324-f19e82c4f3e5.png">

I end up with this result:

<img width="340" alt="screenshot 2019-02-14 at 08 23 24" src="https://user-images.githubusercontent.com/383250/52769865-e250fd00-3031-11e9-93cd-e6fdf232106f.png">

If I disable `g:LanguageClient_completionPreferTextEdit`, things work as expected.

dimbleby:
I believe that the specific way in which prefer-text-edits is broken is per my comment in https://github.com/autozimu/LanguageClient-neovim/issues/857#issuecomment-517236481

> The way that this language client tries to implement prefer-text-edits is to accept the completion as specified by the server in the label, and then when the completion is done it undoes that and applies the text edits instead.
> 
> See languageClient_handleCompleteDone, called at the CompleteDone autocmd - here is the undo.
> 
[...]

PR #858, opened 2019-08-01
Better handling of text-edits at completion

dimbleby:
The current implementation calls `undo`, but unless you happened to break the undo sequence before starting the completion - unlikely in regular typing - this will go very wrong:

- it'll undo more than just the insertion of the unwanted word (potentially a lot more)
- and having done that, it'll perform the text edit in the wrong place in the document

The approach here is borrowed from https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/omni.vim#L321

- assume that the text edit is taking place at the completion point
- tweak that text edit so that it removes the now-unwanted completion

cf #662, https://github.com/autozimu/LanguageClient-neovim/issues/857#issuecomment-517236481

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -2431,8 +2431,20 @@ impl LanguageClient {
         let mut edits = vec![];
         if self.get(|state| state.completionPreferTextEdit)? {
             if let Some(edit) = lspitem.text_edit {
-                self.vim()?.command("undo")?;
-                edits.push(edit.clone());
+                // The text edit should be at the completion point, and deleting the partial text
+                // that the user had typed when the language server provided the completion.
+                //
+                // We want to tweak the edit so that it instead deletes the completion that we've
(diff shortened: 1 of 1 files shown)


src/language_server_protocol.rs
```diff
@@ -2431,8 +2431,20 @@ impl LanguageClient {
...
-                edits.push(edit.clone());
+                // The text edit should be at the completion point, and deleting the partial text
+                // that the user had typed when the language server provided the completion.
+                //
+                // We want to tweak the edit so that it instead deletes the completion that we've
+                // already inserted.
+                //
+                // Check that we're not doing anything stupid before going ahead with this.
+                let mut edit = edit.clone();
+                edit.range.end.character =
```
autozimu, line 2442: This is smart!

I will test a bit more before merging this in.

autozimu:
Thanks!

PR #862, opened 2019-08-07
rangeFormatting does not properly transmit range

afflux:
rangeFormatting does not work for me, as the parameters use a different syntax when parsed in the rust client: https://github.com/autozimu/LanguageClient-neovim/blob/next/src/language_server_protocol.rs#L1455

Fix attached


autozimu:
Thanks!

Issue #864, opened 2019-08-09
Hover floating window auto-close breaks visual selection


YaLTeR:
## Describe the bug
Starting a visual selection while the floating window hover is active and then moving the cursor stops the visual selection. This is especially annoying when using any kind of automatic hover—in this case you must actively attempt to start the selection with the floating window closed.

## Environment
- neovim/vim version (`nvim --version` or `vim --version`): `NVIM v0.4.0-1587-g1d6e36815`
- This plugin version (`git rev-parse --short HEAD`): 2c9c091
- This plugin's binary version (`bin/languageclient --version`): `languageclient 0.1.147`
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue. Refer to an example [here][min-vimrc.vim]): `min-vimrc.vim`
- Language server link and version: RLS or `rust-analyzer` or whatever

[min-vimrc.vim]: https://github.com/autozimu/LanguageClient-neovim/blob/next/min-vimrc.vim

## To Reproduce
1. `cargo new test-project`
1. `nvim -u min-vimrc.vim test-project/src/main.rs`
1. Add `/// a` above `fn main()` so that it gives more than one line of hover.
1. `:call LanguageClient_textDocument_hover()` while on `main`.
1. <kbd>v</kbd> or <kbd>V</kbd> or <kbd>Ctrl</kbd>+<kbd>v</kbd> to start the visual selection and try to move the cursor.
1. Observe how the hover window disappeared and the visual selection was cleared.

## Current behavior
The visual selection is cleared.

## Expected behavior
The visual selection is not cleared.

autozimu:
Shall fixed by #866

YaLTeR:
Huh? That's related?

autozimu:
Oops, #865

PR #865, opened 2019-08-11
 Add LanguageClient_closeFloatingHover() 

YaLTeR:
My use-cases:

- I have automatic hover on CursorMove, and due to the order of autocmds, it's triggering before LanguageClient's autocmd has a chance to close the previous hover window, effectively only opening hover every other cursor move. Now I can close it manually before it's spawned again.

- Starting a visual selection while the hover window is shown and moving the cursor clears the visual selection (#864). By mapping visual selection keys to close the hover window right before starting the visual selection, that issue is effectively avoided.


autozimu:
Thanks!

PR #867, opened 2019-08-14
restore goto jdt content feature

cheuk-fung:
Previously `java_classFileContents` function only request content from
language server, now it also opens a new buffer with the content.


autozimu:
Thanks!

PR #869, opened 2019-08-18
Send messages for jdt files

cheuk-fung:
JDT files are essentially Java source files. By sending messages for jdt
files, we are able to use language server functionalities in those
files, such as go to definition or type definition which is supper
helpful for browsing source code.


autozimu:
Thanks!

Issue #871, opened 2019-08-22
Support for a customizable virtual text prefix string


ghost:
## Is your feature request related to a problem? Please describe.
While working on files having multiple errors, warnings and hints (etc.) with virtual text enabled, it can sometimes get difficult to read and focus on the code. Striking the balance between applying a subtle highlight to virtual text and still keeping it visible enough to be really useful can be challenging.

## Describe the solution you'd like
I propose that an option for a customizable virtual text prefix should be added to the plugin.

## Describe alternatives you've considered
Customizing the highlighting for the virtual text can make the readability better, adding white space between normal text and virtual text further improves it, especially if you have weak eyesight.

## Additional context
Blow are some examples of what it can look like:

Image with no virtual text prefix:
![no-prefix](https://user-images.githubusercontent.com/36127183/63550580-68ec1700-c54c-11e9-91bb-e76301ba1d97.png)

Image with virtual text prefix set to `'    •••➜ '`
![prefix](https://user-images.githubusercontent.com/36127183/63550630-77d2c980-c54c-11e9-91d8-ba28c1ddb046.png)
Adding symbols to the prefix with or without white space makes it very clear where the text ends and the virtual text begins.

There is pull request accompanying this issue which implements the proposed solution.
Thank you.

autozimu:
Already addressed in #872

PR #872, opened 2019-08-22
Added customizable virtual text prefix

ghost:
Hi,
This pull request adds a new feature for a customizable virtual text prefix, which can be used to improve readability of text (by adding white space between text and virtual text) or generally for aesthetic customization. It introduces a new global variable `g:LanguageClient_virtualTextPrefix` which can be used to set the virtual text prefix. By default it is initialized to an empty string so existing users of the plugin will not observe any changes after updating. 

This is what the defaults look like:
no-prefix

And this is whith `g:LanguageClient_virtualTextPrefix = '    •••➜ '`
prefix

This pull request also contributes documentation for the new feature.

This pull request addresses the issue: #871 

I am new to this, if there are any problems please let me know.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -148,6 +148,11 @@ endfunction
 function! s:set_virtual_texts(buf_id, ns_id, line_start, line_end, virtual_texts) abort
     " VirtualText: map with keys line, text and hl_group.
 
+    let l:prefix = s:GetVar('LanguageClient_virtualTextPrefix')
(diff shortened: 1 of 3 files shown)


autozimu:
Awesome! Thanks.

ghost:
No problem and thank you for the plugin, It has so far provided me with the best LS experience in neovim. And is consistently providing better completions and user experience then coc.vim (really loving the automatic quickfix list population).

PR #874, opened 2019-08-24
Fix error in FZFSinkLocation when filename contains ':'

cheuk-fung:
In function languageClient_FZFSinkLocation, ':' is used to split the
selected location to get tokens for [filename, line, character]. But if
filename contains ':', the length of tokens will be greater than 3. This
fix joins tokens[..(len - 2)] back with ':' to form the correct token
vector [filename, line, character].

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -287,7 +287,7 @@ impl LanguageClient {
                 self.apply_TextEdits(&uri.filepath()?, edits)?;
             }
         }
-        self.vim()?.edit(&None, &filename)?;
+        self.edit(&None, &filename)?;
         self.vim()?
             .cursor(position.line + 1, position.character + 1)?;
         debug!("End apply WorkspaceEdit");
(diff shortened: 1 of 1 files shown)


src/language_server_protocol.rs
```diff
@@ -894,6 +894,16 @@ impl LanguageClient {
         Ok(())
     }
 
+    fn edit(&self, goto_cmd: &Option<String>, path: impl AsRef<Path>) -> Fallible<()> {
+        let path = path.as_ref().to_string_lossy();
+        if path.starts_with("jdt://") {
```
autozimu, line 899: What is the reason for this change? 

I assume it should be addressed in https://github.com/autozimu/LanguageClient-neovim/pull/867

cheuk-fung: #867 works only if `locations` contain one element. If there are more elements in `locations`, selection ui will pop up. For example, on a Java interface (eg `List`) that has multiple implementations, call `LanguageClient#textDocument_implementation()` will pop up the selection ui.

For FZF, the `FZFSinkLocation` function will be called after user selects a location. Unfortunately, two issues in `FZFSinkLocation` will prevent displaying jdt content:
- jdt uri contains `:`, which is addressed in the pull request description
- `FZFSinkLocation` calls `self.vim()?.edit()` to load jdt uri, same issue as #867. `self.edit()` function is added to fix this issue. All invocation to `self.vim()?.edit()` is updated to use this function instead so jdt uri can be handled correctly for all use cases.

I hope that answered your question. Thanks.

autozimu: 👍

Issue #427, opened 2018-05-13
Completion with items containing dots.


randy3k:
## Summary

In the R language, variable names could contain dots, e.g., `foo.bar` is one single variable.

Suppose I typed `file.c` in R, the possible completions are `file.create`, `file.choose` and `file.copy`.
However, when I hit `C-x, C-o`, I got this

<img width="319" alt="screen shot 2018-05-13 at 12 59 58 am" src="https://user-images.githubusercontent.com/1690993/39964027-f8a5fe16-5648-11e8-8abf-4f49290713c0.png">

The completion is inserted at the dot but not at the first `f`. I tried setting `iskeyword` but it doesn't help.

## Reproduction
- neovim/vim version (`nvim --version` or `vim --version`):
neovim v0.2.2
- This plugin version (`git rev-parse --short HEAD`):
948677f
- This plugin's binary version (`bin/languageclient --version`):
languageclient 0.1.73 bfa74f1254805b463dbd1cbe0f24fa4eee0ec866
- Minimal vimrc content. A minimal vimrc is the smallest vimrc that could
```
set completefunc=LanguageClient#complete
set formatexpr=LanguageClient#textDocument_rangeFormatting()
let g:LanguageClient_autoStart = 1
let g:LanguageClient_diagnosticsEnable = 1
let g:LanguageClient_serverCommands = {}
let g:LanguageClient_serverCommands.r = ['R', '--quiet', '--slave', '-e', 'languageserver::run(debug = TRUE)']
```
- Language server link and version:
https://github.com/REditorSupport/languageserver
v0.2.2

PR #876, opened 2019-09-03
Fix deoplete unicode completion

pappasam:
* Passes the tests

tl;dr LanguageClient-neovim's Deoplete source caused completion failures when completing strings AFTER unicode characters had been entered earlier on the same line. This PR resolves the issue.

## Symptoms

I spent the weekend writing a language server for a writing project that I used with LanguageClient-neovim and Deoplete. This language server supplies me with string values with lots of Unicode special characters, which presents an interesting edge case for this LanguageClient.

Things worked fine as long as a line began with ASCII-like characters, but broke down once special unicode characters were present:

```txt
This line is totally fine, and words complete through the end :)

This līōn will stop completing correctly after the first līōn :(
```

## Cause

I'm not 100% certain about what was happening, but my strong hunch: VimL string functions historically operated on Bytes. The get_complete_position function implementation referenced a vimscript function that uses 'match', which is one-such function. See below:

```viml
function! LanguageClient#get_complete_start(input) abort
    " echomsg a:input
    return match(a:input, '\k*$')
endfunction
```

For ASCII-like strings, bytes can be treated identically to Vim characters. When we get to the broader Unicode universe, bytes and strings behave differently in VimL.

## Fix

Delete the overriding method.

Luckily, the Deoplete plugin provides a default implementation for get_complete_position that I believe is perfectly suitable for our use-cases. No need to re-invent the wheel unless we have to!

## Additional work

I did some minor tweaks to the Deoplete plugin, cleaning up the code. I've separated the changes into their own commits; please let me know if anything is unclear.

### DIFF
--- a/rplugin/python3/deoplete/sources/LanguageClientSource.py
+++ b/rplugin/python3/deoplete/sources/LanguageClientSource.py
@@ -1,7 +1,7 @@
 from .base import Base
 
 
-CompleteOutputs = "g:LanguageClient_omniCompleteResults"
+COMPLETE_OUTPUTS = "g:LanguageClient_omniCompleteResults"
 
 
 class Source(Base):
@@ -18,8 +18,8 @@ class Source(Base):
 
     def gather_candidates(self, context):
         if context["is_async"]:
-            outputs = self.vim.eval(CompleteOutputs)
-            if len(outputs) != 0:
+            outputs = self.vim.eval(COMPLETE_OUTPUTS)
+            if outputs:
                 context["is_async"] = False
                 # TODO: error handling.
                 candidates = outputs[0].get("result", [])
@@ -27,7 +27,7 @@ class Source(Base):
                 return candidates
         else:
             context["is_async"] = True
-            self.vim.command("let {} = []".format(CompleteOutputs))
+            self.vim.command("let {} = []".format(COMPLETE_OUTPUTS))
             character = (context["complete_position"]
                          + len(context["complete_str"]))
             self.vim.funcs.LanguageClient_omniComplete({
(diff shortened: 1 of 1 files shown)


autozimu:
Have you verified this new change won't affect previous closed issues? namely,
- https://github.com/autozimu/LanguageClient-neovim/issues/427
- https://github.com/autozimu/LanguageClient-neovim/pull/456

pappasam:
tl;dr I've manually verified that this new change won't affect the aforementioned previously closed issues.

I'm using my branch on the daily for Python / for my custom language server and have experienced no issues between the two.

#456 I tested by adding non-keyword characters to my language server; I observe no regressions to this issue. Additionally, the PR associated with issue 456 only made changes to the omnifunc implementation, LanguageClient#complete, which is separate from the deoplete fix I made.

#427 I installed the R language server and tested some examples with "." in variable names using both my branch and the current "next" branch. Neither branch experiences the issues with 427. My configuration:

```viml
\ 'r': ['R', '--slave', '-e', 'languageserver::run()'],
```

NOTE: I may have found an existing issue with R: with the R language server configured, Deoplete works fine for me, but omnicomplete (C-x C-o) works in neither the "next" branch nor the branch in this PR. Therefore, desired behavior from #427 works under deoplete but NOT with the omnifunc. Omnifunc works for for Python and my custom language server, so I imagine that 427 may be yet unfixed. A task for another day...

autozimu:
Thank you so much for verifying!

Actually the original PR introduced `get_complete_position` override is https://github.com/autozimu/LanguageClient-neovim/pull/339

Can you check this change doesn't introduce regression regarding https://github.com/autozimu/LanguageClient-neovim/pull/339

autozimu:
Never mind. I verified #339 is not regressed with this change.

autozimu:
Thanks for contributing!

PR #881, opened 2019-09-14
Add a clients_mutex for starting new language servers

### LINKED ISSUE #878: Multiple language servers get started (and terminated) if vim is started with `vim -p`
- Did you upgrade to latest plugin version? yes
- Did you upgrade to/compile latest binary? `languageclient 0.1.147`

## Describe the bug
Multiple php language servers are started up on vim startup when vim opens multiple php files with `vim -p`. This requires a lot of memory to initialize (causes swapping if there are dozens of files), and all but one don't end up being used

I believe this is a race condition in src/language_server_protocol.rs in languageClient_startServer.

The project that I'm editing with vim is very large, so the language server takes more than a second to respond to the language server `initialize` RPC request

```rust
        if self.get(|state| state.clients.contains_key(&Some(languageId.clone())))? {
            return Ok(json!({}));
        }
// much, much later, without a mutex getting acquired globally or for the language id
        self.update(|state| {
            state.clients.insert(Some(languageId.clone()), client);
            Ok(())
        })?;
```
## Environment
- vim 8.1-1-1778
- This plugin version (`git rev-parse --short HEAD`): c851073
- This plugin's binary version (`bin/languageclient --version`):  0.1.147
- Minimal vimrc content: TODO
- Language server link and version: https://github.com/TysonAndre/LanguageServer-phan-neovim (latest)

## To Reproduce
1. Set up phan for a large project and install the language server
1. Start vim, `vim -p file1.php file2.php file3.php file4.php`
1. `ps -ef|grep -i php`. Observe that initially, multiple language servers are running, but afterwards, there are defunct php processes with the `bin/languageclient` as the parent process.

## Current behavior
Multiple php language servers are started up on vim startup when vim opens multiple php files with `vim -p`. This requires a lot of memory to initialize (causes swapping if there are dozens of files), and all but one don't end up being used

## Expected behavior

A single php language server is started up (i.e. one per language, or one per server command in the future)

TysonAndre:
This prevents more than one language server from starting up at any
point in time.

Fixes #878

### DIFF
--- a/src/language_client.rs
+++ b/src/language_client.rs
@@ -2,15 +2,40 @@ use super::*;
 use crate::vim::Vim;
 use std::ops::DerefMut;
 
-pub struct LanguageClient(pub Arc<Mutex<State>>);
+pub struct LanguageClient {
+    pub state_mutex: Arc<Mutex<State>>,
+    pub clients_mutex: Arc<Mutex<HashMap<LanguageId, Arc<Mutex<()>>>>>,
+}
 
 impl LanguageClient {
     // NOTE: Don't expose this as public.
     // MutexGuard could easily halt the program when one guard is not released immediately after use.
     fn lock(&self) -> Fallible<MutexGuard<State>> {
-        self.0
+        self.state_mutex
             .lock()
             .map_err(|err| format_err!("Failed to lock state: {:?}", err))
+    }
+
+    // This fetches a mutex that is unique to the provided languageId.
+    //
+    // Here, we return a mutex instead of the mutex guard because we need to satisfy the borrow
+    // checker. Otherwise, there is no way to guarantee that the mutex in the hash map wouldn't be
+    // garbage collected as a result of another modification updating the hash map, while something was holding the lock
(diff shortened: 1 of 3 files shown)


src/language_client.rs
```diff
@@ -2,13 +2,22 @@ use super::*;
...
-        self.0
+        self.state_mutex
+            .lock()
+            .map_err(|err| format_err!("Failed to lock state: {:?}", err))
+    }
+
+    pub fn lock_clients_update(&self) -> Fallible<MutexGuard<()>> {
+        self.clients_mutex
             .lock()
             .map_err(|err| format_err!("Failed to lock state: {:?}", err))
```
autozimu: This error message shall be update to `... lock client creation`

autozimu:
Can you upgrade to take languageId into consideration? Say user invoke vim with `vim -p file1.php file2.java`, then this change would fail one of the server startup.

Sorry didn't bring up in the first contact.

TysonAndre:
> Can you upgrade to take languageId into consideration? Say user invoke vim with `vim -p file1.php file2.java`, then this change would fail one of the server startup.

The review comments have been addressed, this is ready for another look

autozimu:
Thanks!

PR #883, opened 2019-09-16
Fix install fails

hrfmmr:
Because the built binary download URL is different from that of [releases](https://github.com/autozimu/LanguageClient-neovim/releases)

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,6 +1,6 @@
 #!/usr/bin/env pwsh
 
-$version = '0.1.153'
+$version = 'v0.1.153'
 $name = 'languageclient'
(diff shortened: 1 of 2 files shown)


autozimu:
Thank you so much for reporting!

As documented in reference commit, the root cause of this issue is that cargo-release starts to use `v` as prefix starting version 0.12.0 https://github.com/sunng87/cargo-release/blob/master/CHANGELOG.md#changed-4

The same reference commit shall make cargo release behave identically as previously.

I will also publish a new binary release shortly.

hrfmmr:
Got it, thank you!

PR #885, opened 2019-09-20
Fix: install compare with the binary's version

jsvisa:
the `installed_version` should be fetched from the binary, not the given one.


autozimu:
Thanks!

PR #895, opened 2019-09-26
Show relatedInformation in explainErrorAtPoint

jez:
The LSP spec allows for relatedInformation on a diagnostic, which is how
VS Code shows error messages that look like this:

Screen Shot 2019-09-25 at 6 30 11 PM

This change implements similar logic for explainErrorAtPoint by looping
over the relatedInformation fields and concatenating them together.
Here's what it looks like in my Vim setup:

Screen Shot 2019-09-26 at 10 13 45 AM

Before this change, it looked like this:

Screen Shot 2019-09-25 at 6 35 17 PM


autozimu:
Thanks!

PR #896, opened 2019-09-26
Add hideVirtualTextsOnInsert option

### LINKED ISSUE #816: Feature Request: an option to disable notifications in insert mode
I am disturbed by the error messages and hints when I'm editing my code and not finished yet.

Provide an option to disable these notifications in insert mode, and when I'm switching to normal mode, display them again.

Consider to make the above being default behavior.

martskins:
This pull request fixes #816 by adding an option (`LanguageClient_hideVirtualTextsOnInsert`) to hide diagnostics in virtual texts while typing (insert mode) and setting `1` as a default value for it.

Here's what the end result looks like:

recording-2019-09-26_07 57 56

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -374,6 +374,13 @@ Maximum severity to show diagnostic messages.
 Default: "Hint"
 Valid options: "Error" | "Warning" | "Information" | "Hint"
 
+2.29 g:LanguageClient_hideVirtualTextsOnInsert *g:LanguageClient_hideVirtualTextsOnInsert*
+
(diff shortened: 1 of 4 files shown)


martskins:
On thing to note is that this causes the diagnostics not to show in virtual texts from the moment you stop editing and exit insert mode until the moment you move the cursor. This is because the call to `set_virtual_texts` in `didChange` clears the virtual texts but only a `handleCursorMoved` places them back, as `set_virtual_texts` is not called in `didSave`.

How do you feel about adding a call to `set_virtual_texts` in `didSave`?

martskins:
Closing in favor of #982

PR #900, opened 2019-10-14
Fix WorkspaceEditWithCursor definition

martskins:
I'm not sure if there's an open issue for this, but while working with rust-analyzer I got some errors while trying to run the `codeAction` command. Basically, I would get something like this when trying to run a codeAction that modified the current buffer:

Screenshot from 2019-10-14 18-33-48

Digging around I found that the definition of `WorkspaceEditWithCursor` expects a `TextDocumentPositionParams` in the field `cursorPosition`, but as I understand it, that field [should be optional](https://github.com/rust-analyzer/rust-analyzer/blob/05ed6c548a7672e2c9472276a652c374a5d2a212/editors/code/src/commands/apply_source_change.ts#L9).

This PR fixes the definition of `WorkspaceEditWithCursor`, making the `cursorPosition` optional, to properly handle the `applySourceChange` command from rust-analyzer.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -841,10 +841,12 @@ impl LanguageClient {
                     for edit in edits {
                         let edit: WorkspaceEditWithCursor = serde_json::from_value(edit.clone())?;
                         self.apply_WorkspaceEdit(&edit.workspaceEdit)?;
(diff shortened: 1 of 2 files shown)


autozimu:
Thanks!

PR #901, opened 2019-10-15
Add installation for dein

ekremkaraca:


autozimu:
This is already in https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md

It wasn't added in README is that I am trying to keep it short and clean. User shall read INSTALL to find complete instructions.

Issue #763, opened 2019-03-10
Limit Diagnostics to the currently focused buffer


martsa1:
## Is your feature request related to a problem? Please describe.
Currently, the diagnostics output displays diagnostics from all open buffers.  On even moderate projects, this means that information relevant to the current buffer is quickly lost.

## Describe the solution you'd like
It would be great to be able to configure diagnostics output to show either the current buffer only, or preserve the current behaviour of showing diagnostics from all open buffers.

## Describe alternatives you've considered
#726 would potentially allow me to configure filtering based on the current buffer's file name which would achieve pretty much what I want.

The only other alternative is to edit only one or two files at a time and restart neovim regularly, which doesn't seem reasonable, especially as neomake + pylint + flake8 behaves the way I would expect out of the box.

## Additional context

Sample of what I currently see.
![image](https://user-images.githubusercontent.com/3308997/54088795-a7549780-4359-11e9-9e90-de55b7830f1a.png)

From what I can make of the pyls debug output, it appears to correctly send only diagnostics for the one file from the didChange event etc, so I'm guessing that the language client is holding onto previous diagnostics?

Neovim Version: 0.3.1
LanguageClient-neovim version: Up-to-date
Python-Language-Server version: Up-to-date

Thanks for all your work and effort with the language client, its amazing!

autozimu:
The behaviour of diagnostic list is modelled after vim quickfix, which shows all related issues of the project regardless of which files is currently being opened, since the list is global. The current behaviour is suggested by another user. So I guess it's best to keep the current behaviour unless there is a compelling reason not to.

BTW, you should be able to configure the linter to show only error levels.

martsa1:
I would politely disagree with your perspective here, the codebase I spend most of my day working in consists of > 5000 python files (and I'll regularly have > 30 open at once).  If the diagnostics are designed to behave on a project-wide manner, I'd expect diagnostics from all 5000 odd python files, rather than just the ones I've touched, which would create potentially a large overhead when running through the various diagnostics?

Personally, I'd prefer to see diagnostics related only to my current buffer.  I have also configured my client to output diagnostics to the location list, as I regularly use the quickfix for find and replace.

Whilst configuring the client to filter by error level would help reduce the cruft to the key information, I'd regularly want to change the level its set to.  Perhaps thats possible with some vimscript (I really need to get to grips with that), but I feel a config option for the language client to show per current buffer vs all open buffers (the current behaviour) would be a useful option.

[...]

PR #908, opened 2019-10-30
Clear previous diagnostics in case if LanguageClient_diagnosticsList is "Location"

ruslan-savina:
If g:LanguageClient_diagnosticsList = 'Location' we have to clear previous diagnostic results.
As a result, we will have diagnostics only for a current buffer in case if 'Location' is set.

Fixes:
#763

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1941,7 +1941,7 @@ impl LanguageClient {
             match state.diagnosticsList {
                 DiagnosticsList::Location => {
                     state.diagnostics.clear();
-                },
+                }
                 _ => {}
             };
             state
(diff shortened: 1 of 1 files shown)


autozimu:
It seems like this would depend on language server's publishing diagnostics sequence. It only works when language server publishes diagnostics for current buffer last.

ruslan-savina:
I've checked the [specification](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/) and it looks like there is no such thing as diagnostics publishing sequence. All language servers should work identically.

autozimu:
> All language servers should work identically

There is no guarantee this is true.

A better way would be comparing current buffer filename with the one in diagnostic list.

autozimu (changes requested):
Please check previous comment

Issue #916, opened 2019-11-22
Flag for disabling diagnostics sorting


tdroxler:
## Is your feature request related to a problem? Please describe.

I noticed that the diagnostics I send from my server were not in the same order in my quickfix list. I found here: https://github.com/autozimu/LanguageClient-neovim/blob/next/src/language_server_protocol.rs#L1926 that the diagnostics are now sorted. I'm wondering if it's the role of the client to sort them? Anyway I suppose you had good reasons to do it, so I was thinking of a flag to disable the sorting.

## Describe the solution you'd like

Having a flag to enable/disable the sorting of the diagnostics, to be able to have them in the same order as sent by the server.

## Describe alternatives you've considered
I have currently sadly moved back to a prior commit 

Thx a lot, cheers

autozimu:
The reason diagnostics list is sorted is that one line might have multiple diagnostics and I figured high severity diagnostics should triumph lower ones during highlighting.

The problem here is that I probably got it wrong about when to kick off this sorting. You're right that the client shall preserve the diagnostic list order while presenting in quickfix etc.

autozimu:
I will see if I can get time to create a fix soon.

tdroxler:
Thx a lot

PR #918, opened 2019-11-25
Add a flag to toggle diagnostics sorting

tdroxler:
Hi,

as discussed in #916 it's sometime useful to keep the order of diagnostics as send by the server.

First time I do some rust, so I not sure I did everything right.


autozimu:
Closing. See https://github.com/autozimu/LanguageClient-neovim/issues/916#issuecomment-560227061

PR #920, opened 2019-11-29
Download binaries from `vim`

abitrolly:
`:LanguageClientInstallBinaries` makes it easy to download or refresh `languageclient`. This is my first time writing `vimscript` and I just wrapped provided `install.sh`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -64,21 +64,14 @@ Restart neovim and run `:call dein#install()` to install.
 
 ## Manual
 
-Clone this repo into some place, e.g., `~/.vim-plugins`
+Clone this repo into `vim` plugin dir.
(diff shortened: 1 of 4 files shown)


autozimu:
What are the benefits of providing this extra command than relying on plugin manager like plug or dein to do the work automatically? If the user needs to do the download one time occasionally, it should be easy enough to fire off the command in terminal.

abitrolly:
@autozimu I don't use plugin managers with Vim 8, because it is able to discover and run plugin itself. The ability to download/refresh binaries is a convenience feature I used to in `vim-go`. Instead of directing users to RTFM in case of missing binaries, it offers a quick way to fix problem with one command.

PR #928, opened 2019-12-20
Properly show signs on last line of viewport

TysonAndre:
`:echo line('w$')` will echo the line number **of the last visible**
line, not the line number after the last visible line.
(this can be seen by invoking this manually with line numbers on)

So this should be checking `<=` to be true for the last line number.

```
autoload/LSP.vim
function! LSP#viewport() abort
    return {
        \ 'start': line('w0') - 1,
        \ 'end': line('w$'),
        \ }
endfunction
```

**EDIT: The language server I'm using emits diagnostics with start: (line N, column 0), end: (line N+1, column 0). But that doesn't seem like it'd cause bugs in overlaps, anyway**

### DIFF
--- a/src/viewport.rs
+++ b/src/viewport.rs
@@ -16,7 +16,7 @@ impl Viewport {
     }
 
     fn contains(&self, line: u64) -> bool {
-        line >= self.start && line < self.end
+        line >= self.start && line <= self.end
     }
 
     pub fn overlaps(&self, range: Range) -> bool {


autozimu:
Is this causing any issue?

I believe here both `start` and `end` are 0-base with start inclusive and end exclusive. So the original implementation looks fine to me.

TysonAndre:
It seems like my issues were caused by a race condition in updating (I may have been testing with a debug release).

Signs do show up, usually after I move the cursor up (with `k` in normal mode)

And looking again, all the lines involved are 0-based

TysonAndre:
> It seems like my issues were caused by a race condition in updating (I may have been testing with a debug release).

Also, the issues I was seeing were on my fork, with the `sign unplace * file=...` workaround, so it's probably a race condition from that.

PR #929, opened 2019-12-20
Remove crossbeam-channel in favor of crossbeam

martskins:
Remove crossbeam-channel, as it is no longer maintained, and replace it with crossbeam.


autozimu:
Thanks!

PR #930, opened 2019-12-20
Drop `OptionDeref` in favor of the now stable `Option::as_deref`

martskins:
Drop `OptionDeref` in favor of the new `Option::as_deref` stabilised in Rust 1.40


autozimu:
Thanks!

PR #935, opened 2019-12-27
Implement codeLens

martskins:
Still some missing pieces, but thought I'd open a draft PR to get some feedback while I finish it.

Made some decisions that might need revisiting, like changing the type of the value of `LanguageClient_useVirtualText` from an int to a string with one of these values: All, CodeLens, Diagnostics, No. The default for this setting is `CodeLens`, which will show only codeLens in virtual texts, but can be changed to `All` which will show both CodeLens and Diagnostics in virtual text.

It supports running codeLens actions via a new function `LanguageClient#codeLensAction` which runs the action associated with the codeLens under the cursor. Only `rust-analyzer.runSingle` and `rust-analyzer.run` are handled now, which correspond to run a single test and run a binary respectively, and they are handled by running the corresponding command in a `neovim` terminal (`vim8` approach yet to be defined).

Not sure what to do about vim, as it doesn't support virtual text (afaik), and I doubt people are going to be triggering codeLens actions without visual feedback on where they are.

Here's what it looks like now:
 
image

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -122,7 +122,11 @@ function! s:useVirtualText() abort
         return l:use
     endif
 
-    return exists('*nvim_buf_set_virtual_text')
+    if exists('*nvim_buf_set_virtual_text')
+        return 'All'
+    else
+        return 'No'
+    endif
 endfunction
 
 function! s:IsTrue(v) abort
@@ -1296,7 +1300,7 @@ function! LanguageClient#java_classFileContents(...) abort
     return LanguageClient#Call('java/classFileContents', l:params, l:Callback)
 endfunction
 
-function! LanguageClient#codeLensAction(...) abort
+function! LanguageClient#handleCodeLensAction(...) abort
     let l:Callback = get(a:000, 1, v:null)
     let l:params = {
                 \ 'filename': LSP#filename(),
@@ -1304,7 +1308,7 @@ function! LanguageClient#codeLensAction(...) abort
                 \ 'character': LSP#character(),
                 \ }
     call extend(l:params, get(a:000, 0, {}))
-    return LanguageClient#Call('LanguageClient_CodeLensAction', l:params, l:Callback)
+    return LanguageClient#Call('LanguageClient/handleCodeLensAction', l:params, l:Callback)
 endfunction
 
 function! LanguageClient_contextMenuItems() abort
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -119,7 +119,7 @@ endfunction
 function! s:useVirtualText() abort
     let l:use = s:GetVar('LanguageClient_useVirtualText')
     if l:use isnot v:null
-        return !!l:use
+        return l:use
     endif
 
     return exists('*nvim_buf_set_virtual_text')
@@ -816,11 +816,23 @@ function! LanguageClient#workspace_symbol(...) abort
     return LanguageClient#Call('workspace/symbol', l:params, l:Callback)
 endfunction
 
-function! LanguageClient#textDocument_codeAction(...) abort
+function! LanguageClient#textDocument_codeLens(...) abort
     let l:Callback = get(a:000, 1, v:null)
     let l:params = {
                 \ 'filename': LSP#filename(),
                 \ 'text': LSP#text(),
+                \ 'line': LSP#line(),
+                \ 'character': LSP#character(),
+                \ 'handle': s:IsFalse(l:Callback),
+                \ }
+    call extend(l:params, get(a:000, 0, {}))
+    return LanguageClient#Call('textDocument/codeLens', l:params, l:Callback)
+endfunction
+
+function! LanguageClient#textDocument_codeAction(...) abort
+    let l:Callback = get(a:000, 1, v:null)
+    let l:params = {
+                \ 'filename': LSP#filename(),
                 \ 'line': LSP#line(),
                 \ 'character': LSP#character(),
                 \ 'handle': s:IsFalse(l:Callback),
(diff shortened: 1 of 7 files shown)


theHamsta:
Nice PR!

Maybe you could remove the saved code lenses from the HashMap once a document is closed.

I played a bit around to see how `rust-analyzer` implements those type hints we see in vscode. I made some experiments on top of your PR. But since there seems no good possibility to display them we have to archive them until Neovim implements inline virtual texts: 
https://github.com/autozimu/LanguageClient-neovim/pull/938

martskins:
Removing them doesn't sound bad, but I think diagnostics don't do that either, so I'd rather keep the behavior the same and maybe refactor that bit for both `diagnostics` and `code_lenses` later on?

Re: inlay hints, so that's what they are called! Good to know! Too bad we don't have that yet, it would definitely be nice!

I think it may be safe to move this to ready for review, as I can't think of much more to add here.

doc/LanguageClient.txt
```diff
@@ -344,8 +344,8 @@ Default: >
 
 Specify whether to use virtual text to display diagnostics.
 
-Default: 1 whenever virtual text is supported.
-Valid Options: 1 | 0
+Default: "CodeLens" whenever virtual text is supported.
+Valid Options: "All" | "No" | "CodeLens" | "Diagnostics"
```
autozimu, line 348: Can we use `All` as default here? This will preserve compatibility with previous behavior.

autoload/LanguageClient.vim
```diff
@@ -119,7 +119,7 @@ endfunction
 function! s:useVirtualText() abort
     let l:use = s:GetVar('LanguageClient_useVirtualText')
     if l:use isnot v:null
-        return !!l:use
+        return l:use
     endif
 
     return exists('*nvim_buf_set_virtual_text')
```
autozimu: We need to provide the default setting here. Otherwise when sync setting, there will be an error deserializing.

src/language_server_protocol.rs
```diff
@@ -850,6 +850,18 @@ impl LanguageClient {
                     }
                 }
             }
+            "rust-analyzer.runSingle" | "rust-analyzer.run" => {
+                if let Some(ref args) = cmd.arguments {
+                    let args = args[0].clone();
```
autozimu: Please avoid use `[0]` to get element. If the element doesn't exit, it will be a full stop.

src/language_server_protocol.rs
```diff
@@ -850,6 +850,18 @@ impl LanguageClient {
...
+                    let bin: String =
+                        try_get("bin", &args)?.ok_or_else(|| err_msg("no bin found"))?;
+                    let arguments: Vec<String> = try_get("args", &args)?.unwrap_or_else(|| vec![]);
+                    let cmd = format!("term {} {}", bin, arguments.join(" "));
```
autozimu: Please add at least warning if `term` is not available.

src/language_server_protocol.rs
```diff
@@ -1759,6 +1774,74 @@ impl LanguageClient {
...
     }
 
+    pub fn languageClient_handleCodeLensAction(&self, params: &Value) -> Fallible<Value> {
+        let filename = self.vim()?.get_filename(params)?;
+        let line = self.vim()?.get_position(params)?.line;
+
+        let mut code_lens: Vec<CodeLens> =
+            self.get(|state| state.code_lens.get(filename.as_str()).cloned().unwrap())?;
+        code_lens.retain(|cl| cl.range.start.line == line);
+        if code_lens.len() != 1 {
```
autozimu: Although there is only a small chance code_lens > 1 here, we still have to handle it properly.

src/types.rs
```diff
@@ -32,6 +32,7 @@ pub const REQUEST__NCM2OnComplete: &str = "LanguageClient_NCM2OnComplete";
 pub const REQUEST__ExplainErrorAtPoint: &str = "languageClient/explainErrorAtPoint";
 pub const REQUEST__FindLocations: &str = "languageClient/findLocations";
 pub const REQUEST__DebugInfo: &str = "languageClient/debugInfo";
+pub const REQUEST__CodeLensAction: &str = "LanguageClient_CodeLensAction";
```
autozimu: Please use `/` to keep consistent.

Also, I'd suggest the content to be `LanguageClient/handleCodeLensAction`

doc/LanguageClient.txt
```diff
@@ -477,6 +477,19 @@ Example bindings combining with tpope/vim-abolish:
...
+*LanguageClient_textDocument_codeLens()*
+Signature: LanguageClient#textDocument_codeLens(...)
+
+Computes and displays the codeLens for the currently open file.
+
+*LanguageClient#textDocument_codeLensAction()*
+*LanguageClient_textDocument_codeLensAction()*
+Signature: LanguageClient#textDocument_codeLensAction(...)
+
+Runs the action associated with the codeLens at the current line. If does
```
autozimu: `If` -> `It`

src/language_server_protocol.rs
```diff
@@ -2500,40 +2589,67 @@ impl LanguageClient {
...
-            self.vim()?.set_virtual_texts(
-                bufnr,
-                namespace_id,
-                viewport.start,
-                viewport.end,
-                &virtual_texts,
-            )?;
+
+            for cl in code_lenses {
+                if cl.command.is_none() {
```
autozimu: Nit: use `if Some(...) = ...` to simplify.

blankname: Do you mean `if let Some(...)`?
```
if let Some(command) = cl.command {
	virtual_texts.push(VirtualText {
        line: cl.range.start.line,
        text: command.title,
        hl_group: "Comment".into(),
    });
}
```

autozimu:
Right

src/language_server_protocol.rs
martskins: What would properly be in that case? Maybe trigger the last action (that one should be the one that is showing if there are multiple codeLenses for a single line)? We could eventually refactor that to show a list of actions and let the user decide, but I don't know if that's worth it, because we don't have a way of displaying multiple codelenses for the same line now, and I don't think anyone is going to trigger an action for a codelens they can't see.

autozimu:
How we are handing multiple code lens actions is another topic. But here at
least we need to know this situation happened, either by logging or warn to
user.

martskins:
Thank you for your feedback @autozimu, every comment should be addressed now. Let me know if you find something else.

autozimu:
Thanks!

Marwes:
The arrows show up for me but `LanguageClient#textDocument_codeAction` does not exist for some reason (`LanguageClient#textDocument_codeAction` does however)

martskins:
@Marwes I think you mean something else in one of those `LanguageClient#textDocument_codeAction`s ?

Marwes:
Do you mean that `LanguageClient#codeLensAction` should be called to run the action? That does not exist either

```
:call LanguageClient#codeLensAction

E117: Unknown function: LanguageClient#codeLensAction
```

martskins:
Ah sorry, that was changed to `LanguageClient#handleCodeLensAction` and I seemed to have missed to update it to the documentation.

martskins:
Fixed documentation via #950

PR #937, opened 2019-12-30
Fix codeAction error where it failed to accept `null`

martskins:
Fixes an issue where a valid `textDocument/codeAction` response of `null` would fail to deserialize.

According to the spec, the response of `textDocument/codeAction` is either a list of actions/commands or null (https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_codeAction).

So the response should be deserialized into `Option<CodeLensResponse>` instead of `CodeLensResponse`. This could be seen as a fix to be made in `lsp-types`, but it seems like it was the intended design, as many other responses are treated the same.

Not fixing this causes some errors to display for example with `gopls`, where if no codeActions are available it will return null.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1372,7 +1372,8 @@ impl LanguageClient {
             },
         )?;
 
-        let response: CodeActionResponse = serde_json::from_value(result.clone())?;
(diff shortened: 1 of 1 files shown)


autozimu:
Thanks!

PR #941, opened 2020-01-02
Add highlight symbol under cursor feature

martskins:
First of all, apologies for sending a PR for an unsolicited feature, but I figured I'd try implementing it instead of asking for someone to implement it. If this feature is decidedly not wanted, that's ok, I learned something in the process anyway.

Having made my apology: This PR adds a function, `LanguageClient#highlightSymbolUnderCursor` to highlight references of the symbol under the cursor. It also adds a trigger for `CursorHold` to fire this new function when the user hasn't moved the cursor for `updatetime`.
The feature can be turned off or on by setting the `LanguageClient_highlightSymbolUnderCursor` option, which is set to 1 by default.

The drawback of having this enabled is that it has to call `nvim_buf_clear_highlight` on each `CursorMoved` trigger, as to clear any possible highlighted references caused by this new feature.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -894,6 +894,21 @@ function! LanguageClient#textDocument_rangeFormatting(...) abort
     return LanguageClient#Call('textDocument/rangeFormatting', l:params, l:Callback)
 endfunction
 
+function! LanguageClient#highlightSymbol() abort
+    let l:Callback = get(a:000, 1, v:null)
+    let l:params = {
+                \ 'filename': LSP#filename(),
+                \ 'text': LSP#text(),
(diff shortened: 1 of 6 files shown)


YaLTeR:
But there already is `LanguageClient_textDocument_documentHighlight`?

martskins:
Oh, interesting, I always took that one as a method to highlight the whole document, not just the symbol under the cursor. My bad, I misunderstood that one. Would a config option to set an `autocmd` to trigger `LanguageClient#textDocument_documentHighlight` be something we are interested then? It would require cleanup of the highlighted symbols on `CursorMoved` too. If so, I can rework this PR to use that one instead of the one I introduced unnecessarily.

PR #944, opened 2020-01-06
Fix threading issues in documentHighlight

YaLTeR:
Due to code being split between critical sections, when many documentHighlight requests were processed at once, it was possible (and very easy) for some highlights to not get cleaned up properly and be left hanging in the editor forever. By putting all state reading and updating into a single critical section, this is no longer possible as state updates are now fully coherent.

In particular, I have documentHighlight on CursorMove, and an easy way to trigger the issue is to use rust-analyzer, open a file like this and hold `w`:
```rust
use nom::branch::alt;
use nom::number::complete::recognize_float;
use nom::bytes::complete::tag;
use nom::error::ErrorKind;

fn main() {
    let result = alt((
            recognize_float::<_, (_, ErrorKind)>,
            tag("1esomething"),
    ))("1esomething");

    println!("result = {:#?}", result);
}
```
Since rust-analyzer takes a couple of seconds to initialize, it would handle all generated documentHighlight requests at once, easily triggering the threading issue.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -362,46 +362,51 @@ impl LanguageClient {
 
             let buffer = self.vim()?.get_bufnr(&filename, params)?;
 
-            let source = if let Some(hs) = self.get(|state| state.document_highlight_source)? {
-                if hs.buffer == buffer {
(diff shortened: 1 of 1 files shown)


autozimu:
Thanks!

PR #948, opened 2020-01-08
Check if server has CodeLens capability

martskins:
Check if server has CodeLens capability before issuing the call to textDocument/codeLens

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1829,41 +1829,72 @@ impl LanguageClient {
             return Ok(Value::Null);
         }
 
-        info!("Begin {}", lsp::request::CodeLensRequest::METHOD);
         let filename = self.vim()?.get_filename(params)?;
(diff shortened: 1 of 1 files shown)


src/language_server_protocol.rs
```diff
@@ -1829,41 +1829,72 @@ impl LanguageClient {
...
-        let mut resolved_code_lens = vec![];
-        if let Some(code_lens) = code_lens {
-            for item in code_lens {
-                let mut item = item;
-                if let Some(_d) = &item.data {
-                    if let Some(cl) = client.call(lsp::request::CodeLensResolve::METHOD, &item)? {
-                        item = cl;
+                if capabilities
+                    .code_lens_provider
+                    .unwrap()
```
autozimu, line 1855: Nit: I would avoid using `unwrap` whenever possible except for prototype as it's going to quit the whole process if unwrapping on none.

Anyway, I can merge this in and update myself.

autozimu:
Thanks!

PR #950, opened 2020-01-09
Fix handleCodeLensAction documentation

martskins:
Fixed incorrect documentation about `LanguageClient#handleCodeLensAction`, which was renamed in the process of the CodeLens PR and the documentation was mistakenly left behind.


autozimu:
Thanks!

PR #953, opened 2020-01-16
Add g:LanguageClient_echoProjectRoot to docs

andrewferrier:
This commit adds the `g:LanguageClient_echoProjectRoot` to the docs. I've found this option by digging into the code. For me it's pretty annoying not to be able to disable it, as otherwise the "Project root" messages litter my vim console as I navigate from file to file.

Thanks for your work on this plugin! It's great - and for a moving target and something very new to nvim - very robust!


autozimu:
Thanks!

Issue #383, opened 2018-04-17
Feature Request: Semantic syntax highlighting


ajmwagar:
According to Sourcegraph, most langservers currently support semantic syntax highlighting. 

I am hoping to get syntax highlighting from my ls in vim.

I know that there is an implementation for c++ with this plugin: https://github.com/JBakamovic/cxxd-vim/

I am hoping to see this as a feature for more/all languages. 

Best of luck!

AJ

htfy96:
Just a note: [`vscode-languageserver-node` is proposing their semantic highlighting protocol extension](https://github.com/Microsoft/vscode-languageserver-node/pull/367).

ajmwagar:
Thanks. Good to see progress in this field.

GordianDziwis:
The eclipse java language server already has an [implementation](https://github.com/eclipse/eclipse.jdt.ls/issues/715) based on the [protocol extension proposal](https://github.com/Microsoft/language-server-protocol/issues/513).

marv:
`clangd` also gained support for semantic syntax highlighting based on the proposed protocol extension: https://clang.llvm.org/extra/clangd/Features.html#complete-list-of-features

martskins:
Fixed by #954

PR #954, opened 2020-01-16
Add support for the proposed Semantic Highlighting Protocol

jackguo380:
Mostly done the initial implementation for supporting for the semantic highlighting protocol proposed by people from the vscode language server. As far as I know currently `clangd`  version 9 and `eclipse.jdt.ls` both have working implementations of this protocol.

This PR adds a new setting to configure semantic highlight `g:LanguageClient_semanticHighlightMaps`. Which is a per language map like `serverCommands` of mappings from semantic/textmate scopes to vim highlight groups.

Here are a couple of sample configurations and screenshots to go along.

Using `clangd` 9 for C++
```vim
let g:LanguageClient_semanticHighlightMaps['cpp'] = [
            \ {'Function': ['entity.name.function.cpp']},
            \ {'Function': ['entity.name.function.method.cpp']},
            \ {'CppNamespace': ['entity.name.namespace.cpp']},
            \ {'CppEnumConstant': ['variable.other.enummember.cpp']},
            \ {'CppMemberVariable': ['variable.other.field.cpp']},
            \ {'Type': ['entity.name.type.class.cpp']},
            \ {'Type': ['entity.name.type.enum.cpp']},
            \ {'Type': ['entity.name.type.template.cpp']},
            \ ]

hi! CppEnumConstant ctermfg=Magenta guifg=#AD7FA8 cterm=none gui=none
hi! CppNamespace ctermfg=Yellow guifg=#BBBB00 cterm=none gui=none
hi! CppMemberVariable ctermfg=White guifg=White
```

<span>

</span>
<br><br>

Using `eclipse.jdt.ls` for Java
```vim
let g:LanguageClient_semanticHighlightMaps['java'] = [
            \ {"JavaStaticMemberFunction": ['storage.modifier.static.java', 'entity.name.function.java', '**']},
            \ {"JavaMemberVariable": ['meta.definition.variable.java', 'meta.class.body.java', 'meta.class.java', '**']},
            \ {"Function": ['entity.name.function.java', '**']},
            \ {"Function": ['*', 'entity.name.function.java', '**']},
            \ {"Type": ['entity.name.type.class.java', '**']},
            \ {"Type": ['*', 'entity.name.type.class.java', '**']},
            \ ]

hi! JavaStaticMemberFunction ctermfg=Green cterm=none guifg=Green gui=none
hi! JavaMemberVariable ctermfg=White cterm=italic guifg=White gui=italic
```

<span>

</span>
<br><br>

TODO:
- Support for `vim` 8.1 with `+textprop`
- Disable feature when the version of `vim` or `neovim` does not support the needed APIs
- More tests?

The PR is broken down into 2 commits since I had to upgrade the `lsp-types` crate to get the types needed for semantic highlighting.

Let me know what you think and if there are any improvements/fixes needed!

Related PRs/Issues:
#383
Microsoft/vscode-languageserver-node#367

Also see this document for exactly how the protocol works:
https://github.com/microsoft/vscode-languageserver-node/blob/5a9b33c23de84c3341e011e79221795a8059375b/protocol/src/protocol.semanticHighlighting.proposed.md

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1386,7 +1386,7 @@ function! LanguageClient_contextMenu() abort
     return LanguageClient_handleContextMenuItem(l:options[l:selection - 1])
 endfunction
 
-function! LanguageClient_semanticScopes(...) abort
+function! LanguageClient_showSemanticScopes(...) abort
     let l:params = get(a:000, 0, {})
     let l:Callback = get(a:000, 1, function('s:print_semantic_scopes'))
 
@@ -1420,8 +1420,11 @@ function! LanguageClient#showSemanticHighlightSymbols(...) abort
     return LanguageClient#Call('languageClient/showSemanticHighlightSymbols', l:params, l:Callback)
 endfunction
 
-function! LanguageClient_showCursorSemanticHighlightSymbols() abort
-    return LanguageClient#showSemanticHighlightSymbols({}, function('s:print_cursor_semantic_symbol'))
+function! LanguageClient_showCursorSemanticHighlightSymbols(...) abort
+    let l:params = get(a:000, 0, {})
+    let l:Callback = get(a:000, 1, function('s:print_cursor_semantic_symbol'))
+
+    return LanguageClient#showSemanticHighlightSymbols(l:params, l:Callback)
 endfunction
 
 function! s:print_cursor_semantic_symbol(response) abort
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1393,7 +1393,7 @@ function! LanguageClient_semanticScopes(...) abort
     return LanguageClient#Call('languageClient/semanticScopes', l:params, l:Callback)
 endfunction
 
-function! s:print_semantic_scopes(response)
+function! s:print_semantic_scopes(response) abort
     let l:scope_mappings = a:response.result
 
     let l:msg = ''
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -251,6 +251,16 @@ function! s:MatchDelete(ids) abort
     endfor
 endfunction
 
+function! s:ApplySemanticHighlights(bufnr, ns_id, clears, highlights) abort
+    for clear in a:clears
+        call nvim_buf_clear_namespace(a:bufnr, a:ns_id, clear.line_start, clear.line_end)
+    endfor
+
+    for hl in a:highlights
+        call nvim_buf_add_highlight(a:bufnr, a:ns_id, hl.group, hl.line, hl.character_start, hl.character_end)
+    endfor
+endfunction
+
 " Batch version of nvim_buf_add_highlight
 function! s:AddHighlights(source, highlights) abort
     for hl in a:highlights
@@ -1374,6 +1384,67 @@ function! LanguageClient_contextMenu() abort
     endif
 
     return LanguageClient_handleContextMenuItem(l:options[l:selection - 1])
+endfunction
+
+function! LanguageClient_semanticScopes(...) abort
+    let l:params = get(a:000, 0, {})
+    let l:Callback = get(a:000, 1, function('s:print_semantic_scopes'))
+
+    return LanguageClient#Call('languageClient/semanticScopes', l:params, l:Callback)
+endfunction
+
+function! s:print_semantic_scopes(response)
+    let l:scope_mappings = a:response.result
+
+    let l:msg = ''
+    for mapping in l:scope_mappings
+        let l:msg .= "Highlight Group:\n"
+        let l:msg .= ' ' . l:mapping.hl_group . "\n"
+
+        let l:msg .= "Semantic Scope:\n"
+        let l:spaces = ' '
+        for l:scope_name in l:mapping.scope
+            let l:msg .= l:spaces . l:scope_name . "\n"
+            let l:spaces .= ' '
+        endfor
+        let l:msg .= "\n"
+    endfor
+
+    echo l:msg
+endfunction
+
+function! LanguageClient#showSemanticHighlightSymbols(...) abort
+    let l:params = get(a:000, 0, {})
+    let l:Callback = get(a:000, 1, v:null)
+
+    return LanguageClient#Call('languageClient/showSemanticHighlightSymbols', l:params, l:Callback)
+endfunction
+
+function! LanguageClient_showCursorSemanticHighlightSymbols() abort
+    return LanguageClient#showSemanticHighlightSymbols({}, function('s:print_cursor_semantic_symbol'))
+endfunction
+
+function! s:print_cursor_semantic_symbol(response) abort
+    let l:symbols = a:response.result
+    let l:lines = []
+
+    for symbol in l:symbols
+        if l:symbol.line + 1 == line('.') &&
+                    \ symbol.character_start < col('.') &&
+                    \ col('.') <= symbol.character_end
+            let l:spaces = ''
+            for scope_name in l:symbol.scope
+                call add(l:lines, l:spaces . l:scope_name)
+                let l:spaces .= ' '
+            endfor
+        endif
+    endfor
+
+    if len(l:lines) > 0
+        call s:OpenHoverPreview('SemanticScopes', l:lines, 'text')
+    else
+        call s:Echowarn('No Symbol Under Cursor or No Semantic Highlighting')
+    endif
 endfunction
 
 function! LanguageClient#debugInfo(...) abort
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -400,6 +400,124 @@ the root of a project is detected using g:LanguageClient_rootMarkers.
 Default: 1 to display the messages
 Valid options: 1 | 0
 
+2.30 g:LanguageClient_semanticHighlightMaps   *g:LanguageClient_semanticHighlightMaps*
+
+String to list/map map. Defines the mapping of semantic highlighting "scopes" to
+highlight groups. This depends on the LSP server supporting the proposed
+semantic highlighting protocol, see:
+
+https://github.com/microsoft/language-server-protocol/issues/18
+https://github.com/microsoft/vscode-languageserver-node/issues/368
+
+
+Like |g:LanguageClient_serverCommands| this is a map where the keys are
+filetypes. However the value at each key can be either:
+
+1. A map of highlight group name (see |highlight-groups|) => list of strings:
+>
+    let g:LanguageClient_semanticHighlightMaps = {
+        \ 'java': {
+        \   "Function": ['entity.name.function.java', '**'],
+        \   "Type": ['entity.name.type.class.java', '**'],
+        \ }
+        \ }
+
+2. A list of maps from #1, this is to allow for multiple patterns for the same
+   highlight group.
+>
+    let g:LanguageClient_semanticHighlightMaps = {
+        \ 'java': [
+        \   {"Function": ['entity.name.function.java', '**']},
+        \   {"Type": ['entity.name.type.class.java', '**']},
+        \   {"Function": ['*', 'entity.name.function.java', '**']},
+        \   {"Type": ['*', 'entity.name.type.class.java', '**']},
+        \ ]
+        \ }
+
+The maps associate the highlight group with a semantic scope matching pattern.
+Any symbols that have a scope that matches the pattern will be highlighted
+wtih that highlight group.
+
+There are a fixed set of semantic scopes defined by the LSP server on startup.
+These can be viewed by calling |LanguageClient_semanticScopes| which will
+show all the semantic scopes and their currently mapped highlight group for
+the currently open buffer's filetype.
+>
+     call LanguageClient_semanticScopes()
+
+     == Output from eclipse.jdt.ls ==
+
+     Highlight Group:
+      None
+     Semantic Scope:
+      invalid.deprecated.java
+       meta.class.java
+        source.java
+
+     Highlight Group:
+      None
+     Semantic Scope:
+      variable.other.autoboxing.java
+       meta.method.body.java
+        meta.method.java
+         meta.class.body.java
+          meta.class.java
+           source.java
+
+     ...
+
+Like the patterns, the semantic scope is a list of strings. They are printed
+with increasing indent to make it easier to read. For example the first scope
+is:
+>
+     ['invalid.deprecated.java', 'meta.class.java', 'source.java']
+
+It is currently isn't mapped to any highlight group as indicated by the None.
+
+Often its more useful to find what semantic scope corresponds to a piece of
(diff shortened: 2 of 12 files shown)


YaLTeR:
Oh, this is so cool! Can't wait for this to get merged.

A number of these scopes like `entity.name.function.java` seem to follow the pattern of `entity.name.function.something`, maybe add a couple of default rules mapping all of these to a reasonable `Function`? And the same for any other common scopes.

jackguo380:
Thanks @YaLTeR 

I was considering adding default values but I don't see any good way to do it since the semantic scopes are kind of language server specific. I'm kind of worried that language servers are going to use different scope names or map the scopes differently, so the defaults may look good on one server but look awful on a another one.

I'll probably just add them into the wiki page for each server once this gets merged.

doc/LanguageClient.txt
```diff
@@ -386,6 +386,124 @@ Maximum severity to show diagnostic messages.
...
+        \   {"Function": ['entity.name.function.java', '**']},
+        \   {"Type": ['entity.name.type.class.java', '**']},
+        \   {"Function": ['*', 'entity.name.function.java', '**']},
+        \   {"Type": ['*', 'entity.name.type.class.java', '**']},
+        \ ]
+        \ }
+
+The maps associate the highlight group with a semantic scope matching pattern.
+Any symbols that have a scope that matches the pattern will be highlighted
+wtih that highlight group.
```
autozimu: wtih => with

autoload/LanguageClient.vim
```diff
@@ -1375,6 +1385,67 @@ function! LanguageClient_contextMenu() abort
     return LanguageClient_handleContextMenuItem(l:options[l:selection - 1])
 endfunction
 
+function! LanguageClient_semanticScopes(...) abort
```
autozimu: Let's rename this a bit to `..._show/print` to make the intension more clear.

autoload/LanguageClient.vim
```diff
@@ -1375,6 +1385,67 @@ function! LanguageClient_contextMenu() abort
...
+endfunction
+
+function! LanguageClient#showSemanticHighlightSymbols(...) abort
+    let l:params = get(a:000, 0, {})
+    let l:Callback = get(a:000, 1, v:null)
+
+    return LanguageClient#Call('languageClient/showSemanticHighlightSymbols', l:params, l:Callback)
+endfunction
+
+function! LanguageClient_showCursorSemanticHighlightSymbols() abort
```
autozimu: Suggest `..._showSemanticHighlightUnderCursor`

doc/LanguageClient.txt
```diff
@@ -386,6 +386,124 @@ Maximum severity to show diagnostic messages.
...
+Often its more useful to find what semantic scope corresponds to a piece of
+text. This can be done by calling |LanguageClient_showCursorSemanticHighlightSymbols|
+while hovering over the text of interest.
+>
+     call LanguageClient_showCursorSemanticHighlightSymbols()
+
+In order for a pattern to match a semantic scope the lists must fully match
+in length and list items.
+>
+     ['x', 'y', 'z'] == ['x', 'y', 'z']
```
autozimu: Let's replace the `==` with `matches`  to make it more clear.

src/types.rs
```diff
@@ -438,6 +466,153 @@ impl PartialEq for Highlight {
...
+            arr_c.matches(v),
+        )
+    };
+
+    assert_eq!(do_matches(&t1), (true, true, true, true));
+    assert_eq!(do_matches(&t2), (false, false, false, true));
+    assert_eq!(do_matches(&t3), (false, false, true, true));
+    assert_eq!(do_matches(&t4), (false, true, false, true));
+}
+
```
autozimu: Can you add unit tests covering case of `**`?

src/vimext.rs
```diff
@@ -1,14 +1,17 @@
 use crate::language_client::LanguageClient;
+use crate::types::LCNamespace;
 use failure::Fallible;
 
 impl LanguageClient {
-    pub fn get_or_create_namespace(&self) -> Fallible<i64> {
-        if let Some(namespace_id) = self.get(|state| state.namespace_id)? {
+    pub fn get_or_create_namespace(&self, ns: &LCNamespace) -> Fallible<i64> {
```
autozimu, line 6: Is there some commits missing from here?

I don't see this updated function ever referenced.

doc/LanguageClient.txt
```diff
@@ -386,6 +386,124 @@ Maximum severity to show diagnostic messages.
...
+        \   "Type": ['entity.name.type.class.java', '**'],
+        \ }
+        \ }
+
+2. A list of maps from #1, this is to allow for multiple patterns for the same
+   highlight group.
+>
+    let g:LanguageClient_semanticHighlightMaps = {
+        \ 'java': [
+        \   {"Function": ['entity.name.function.java', '**']},
```
autozimu: Do you think if it's more intuitive if the map is mapping semantic groups to highlight groups?

Another related question, what is the motivation of supporting two styles of this setting? Should one suffice?

autozimu (commented):
Awesome work!

Just left some comments here.

src/vimext.rs
jackguo380: It gets used by the new `textDocument_semanticHighlight` and `languageClient_handleCursorMoved` in `language_server_protocol.rs`. I think github folded the diff for that file due to it being quite long.

doc/LanguageClient.txt
jackguo380: The main issue/reason for the reversed map is that vimscript doesn't support arrays as a key into a dictionary.

Originally I just had a single map but I realized afterward that theres often a need to map multiple scopes to the same highlight group. I guess I could simplify the documentation and only mention the second style but still allow the first if it ever gets used.

doc/LanguageClient.txt
```diff
@@ -386,6 +386,124 @@ Maximum severity to show diagnostic messages.
...
+
+
+Example configuration for eclipse.jdt.ls:
+>
+     let g:LanguageClient_semanticHighlightMaps = {}
+     let g:LanguageClient_semanticHighlightMaps['java'] = [
+                 \ {"JavaStaticMemberFunction": ['storage.modifier.static.java', 'entity.name.function.java', '**']},
+                 \ {"JavaMemberVariable": ['meta.definition.variable.java', 'meta.class.body.java', 'meta.class.java', '**']},
+                 \ {"Function": ['entity.name.function.java', '**']},
+                 \ {"Function": ['*', 'entity.name.function.java', '**']},
```
YaLTeR: Could these two lines be replaced with one `{"Function": ['**', 'entity.name.function.java', '**']}`?

doc/LanguageClient.txt
```diff
@@ -386,6 +386,124 @@ Maximum severity to show diagnostic messages.
...
+     ['x', 'y', '**'] == ['x', 'y', 'z', '1', '2', '3']
+     ['**', '123', '**'] == ['x', 'y', 'z', '123', '1', '2', '3']
+     ['**', 'y', 'z'] != ['x', 'y', 'z', '1', '2', '3']
+
+
+Example configuration for eclipse.jdt.ls:
+>
+     let g:LanguageClient_semanticHighlightMaps = {}
+     let g:LanguageClient_semanticHighlightMaps['java'] = [
+                 \ {"JavaStaticMemberFunction": ['storage.modifier.static.java', 'entity.name.function.java', '**']},
```
YaLTeR: So this list is parsed in order top to bottom until a matching semantic scope is found? What about the first case, in which order is that parsed?

doc/LanguageClient.txt
YaLTeR: Another idea is to represent semantic scopes as a string with some separator, e.g. `storage.modifier.static.java/entity.name.function.java/**`. That would allow using them as keys. Although idk if there are any banned characters from the semantic scope names that would be usable as separators.

src/types.rs
jackguo380: When a `**` gets used it gets converted into the corresponding `SemanticHighlightMatcher` so the string `**` doesn't ever appear in the matcher itself. See `buildSemanticHighlightMatchers` in `language_server_protocol.rs` for the construction.

E.g. an array `['a', 'b', 'c', '**']` would construct the `ArrayStart` matcher.

This is kind of limiting right now since the `**` can only be used at the beginning/end of the array (although it should be sufficient for most use cases). I don't see any good way to evaluate arbitrary patterns of `**` without an overly complex implemenation or hacks like concatentating the items and regexing. Any suggestions?

doc/LanguageClient.txt
jackguo380: I'll add a note about the order. It is from top to bottom.

src/language_server_protocol.rs
```diff
@@ -201,8 +212,48 @@ impl LanguageClient {
...
+        let mut semanticHighlightMaps = HashMap::new();
+
+        for (languageId, v) in userSemanticHighlightMaps {
+            let mut mapping_pairs = Vec::new();
+            match v {
+                Value::Object(mapping) => {
+                    mapping_pairs
+                        .extend(mapping.into_iter().map(|(hl_group, val)| (val, hl_group)));
+                }
+                Value::Array(values) => {
```
autozimu: Let's start with supporting one style first and begin from there.

With either two styles or one style, serde would be able to deserialize json string to the variable, without manually unwrapping and mapping. The added benefits is serde would be able to tell where and why deserialize failed if there is any error.

doc/LanguageClient.txt
autozimu: Let's reduce the number of supported styles here. More styles supported only leads to complicated implementation to maintainer and more confusion to end user.

@YaLTeR cleaver idea! Agreed the tricky part is the choice of separator. As different language have different rules, I don't know there are possible separators working for all languages.

src/vimext.rs
autozimu: Ahh, missed that. Thank you.

src/language_server_protocol.rs
```diff
@@ -818,6 +873,155 @@ impl LanguageClient {
...
+                    .insert(languageId.into(), capability.scopes.unwrap_or_default());
+                Ok(())
+            })?;
+        }
+
+        info!("End parse Semantic Scopes");
+        Ok(())
+    }
+
+    fn buildSemanticHighlightMatchers(
```
autozimu: This can probably be extracted to a utility function.

src/language_server_protocol.rs
```diff
@@ -987,8 +1191,14 @@ impl LanguageClient {
 
         let result: Value = self.get_client(&Some(languageId.clone()))?.call(
             lsp::request::Initialize::METHOD,
+            #[allow(deprecated)]
             InitializeParams {
+                client_info: Some(ClientInfo {
+                    name: "LanguageClient_neovim".into(),
```
autozimu: Let's use "LanguageClient-neovim" instead.

src/language_server_protocol.rs
```diff
@@ -1577,7 +1803,11 @@ impl LanguageClient {
                     tab_size,
                     insert_spaces,
                     properties: HashMap::new(),
+                    trim_trailing_whitespace: None,
+                    insert_final_newline: None,
+                    trim_final_newlines: None,
```
autozimu: Let's use default to fill properties we don't care.

src/language_server_protocol.rs
```diff
@@ -2103,6 +2347,182 @@ impl LanguageClient {
...
+                let mut new_hls = highlights.into_iter().peekable();
+                let semantic_hls = state.semantic_highlights.get_mut(&languageId).unwrap();
+
+                loop {
+                    match (existing_hls.peek(), new_hls.peek()) {
+                        (Some(existing_hl), Some(new_hl)) => {
+                            use std::cmp::Ordering;
+
+                            match existing_hl.line.cmp(&new_hl.line) {
+                                Ordering::Less => {
```
autozimu, line 2415: Instead of comparing each highlight group, can we just delete all highlights on screen and add all newer visible highlights? Similarly to how diagnostics are highlighted. In my opinion, the implementation would be much simplier.

jackguo380: The main issue is that the protocol supports incremental highlighting. So if that case is not handled then it may only apply the changed highlighting.

Currently `eclipse.jdt.ls` sends a single line of highlighting if the line itself was changed, if a line was added/removed the entire document's worth of highlighting is sent.

From this comment (https://github.com/prabirshrestha/vim-lsp/issues/633#issuecomment-573323937) it looks like `clangd 10` has similar behavior except it can also send half the document's worth of highlighting upon add/remove lines.

I'll continue looking at simplifying this logic a bit. Might try a region based approach rather than the current line by line one.

doc/LanguageClient.txt
autozimu: Continuing the same approach presented by @YaLTeR. To solve the issue of ambiguity of separators, a buffer local variable with default can be introduced, say `b:LanguageClient_semanticScopeSeparator` with default of `/`.

e.g., user might have config like
```vim
let g:LanguageClient_semanticHighlightMaps['java'] = {
    \ 'storage.modifier.static.java/entity.name.function.java/.*': 'Function',
}
```

In a different filetype where `/` might be part of the scope themselves,
```vim
autocmd *.elisp setlocal LanguageClient_semanticScopeSeparator='|'
let g:LanguageClient_semanticHighlightMaps['elisp'] = {
    \ 'storage.modifier.static|name.function|.*': 'Function',
}
```

We would construct scope by concatenate the scopes with this separator. The actual highlight group are then being looked up through this map.

IMO, this way is more intuitive and solves the original issue that `**` can only appear in beginning or end.

jackguo380:
@autozimu @YaLTeR Thanks for all the review comments, I just pushed a fairly major change to use regexs rather than `'*'` and `'**'`, please have a look.

In summary, the configuration of `LanguageClient_semanticHighlightMaps` now looks like this:

```vim
let g:LanguageClient_semanticHighlightMaps['java'] = {
        \ '^storage.modifier.static.java:entity.name.function.java': 'JavaStaticMemberFunction',
        \ '^meta.definition.variable.java:meta.class.body.java:meta.class.java': 'JavaMemberVariable',
        \ '^entity.name.function.java': 'Function',
        \ '^[^:]*entity.name.function.java': 'Function',
        \ '^[^:]*entity.name.type.class.java': 'Type',
        \ }
```

The keys of the map are `vim` regexs which are then matched against the semantic scopes concatentated together using the string `LanguageClient_semanticScopeSeparator` which is `:` by default.

E.g.
```vim
['invalid.deprecated.java', 'meta.class.java', 'source.java']
# becomes
'invalid.deprecated.java:meta.class.java:source.java'
```

I figured this is the simplest to implement and most vim users are competent enough in regex to be able to easily use this. A minor implementation defail is that we have to evaluate the regex's in vim due to differences in syntax between vim and rust's regexes.

autozimu:
Awesome! Will take another look when I got a chance.

autozimu:
Thank you very much for the contribution!

PR #955, opened 2020-01-17
Sort diagnostics by line number to prevent stale signs

### LINKED ISSUE #952: signcolumn includes symbols where no diagnostic exists (signcolumn/gutter is stale)
- Did you upgrade to latest plugin version? *Yes*
- Did you upgrade to/compile latest binary? Run shell command 
  `bin/languageclient --version` to get its version number. *Yes, 0.1.156*
- (Neovim users only) Did you check output of `:checkhealth LanguageClient`?

```
                                                                                                                                                                        
  health#LanguageClient#check                                                                                                                                           
  ========================================================================                                                                                              
    - OK: binary found: <...>/.vim/plugged/LanguageClient-neovim/bin/languageclient                                       
    - OK: languageclient 0.1.156 fb02afed15363c8879277b413b89e6ca51589e6a                                                                                               
    - OK: Floating window is supported and will be used for hover
```
- Did you check [troubleshooting]? *Yes* I haven't analysed the log file yet.

[troubleshooting]: https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#6-troubleshooting

## Describe the bug
Some symbols in the gutter / signcolumn remain after the diagnostic is cleared.
 - The quickfix window does not display a diagnostic on the line with the symbol.
 - The status line does not display any diagnostics when the cursor is on the line with the symbol.
 - Toggling the signcolumn does not fix the issue.
 - Redrawing the screen with `ctrl-l` does not fix the issue.
 - Stopping and Starting the language client does not fix the issue.
 - Restarting neovim does fix the issue.
 - I've seen it with every language server I've used (typescript-language-server, javascript-typescript-langserver, eclipse jdtl)

## Environment
```
nvim --version
NVIM v0.4.3
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: <removed>

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```
- This plugin version (`git rev-parse --short HEAD`):
```
fb02afe
```
- This plugin's binary version (`bin/languageclient --version`):
```
languageclient 0.1.156 fb02afed15363c8879277b413b89e6ca51589e6a
```

## To Reproduce
```
call plug#begin('~/.local/share/nvim/plugged')

    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }

call plug#end()

set signcolumn=yes

    \ 'rust': ['npx', 'typescript-language-server', '--stdio'],
    \ }
let g:LanguageClient_loggingLevel = 'INFO'
let g:LanguageClient_virtualTextPrefix = ''
let g:LanguageClient_loggingFile =  expand('~/.local/share/nvim/LanguageClient.log')
let g:LanguageClient_serverStderr = expand('~/.local/share/nvim/LanguageServer.log')
```

everett1992:
Some signs would not be removed from the sign column when their
diagnostic was fixed.

Debugging revealed that signs_to_add would include two signs for the
same line. `state.signs`is keyed by line number so the second sign for
a line will overwrite the first when signs_to_add is recorded in state.
The first sign's id is lost when it's overwritten so it can't be
removed.

signs_to_add included two signs for the same line because itertools
`group_by` only groups consecutive runs (it's more efficient, you only
need to remember the current group and can flush it when the key
changes).

This quick fix sorts diagnostics by line number so that group_by gives
us one group per line number.

fixes #952

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -2536,6 +2536,7 @@ impl LanguageClient {
                         diag.severity.unwrap_or(DiagnosticSeverity::Hint),
                     )
                 })
+                .sorted_by_key(|(line, _)| *line)
                 .group_by(|(line, _)| *line)
                 .into_iter()
                 .filter_map(|(_, group)| group.min_by_key(|(_, severity)| *severity))


everett1992:
There's a couple ways to fix this issue, I chose the easiest, sorting the iterator so group_by works.

I think it would be faster/efficient-er to create a hash map of `line number -> severity` then insert each diagnostic into the map if it's severity is less than the current value at the line number.

Overall I think that state mapping by line number is a bad idea, unless you need to index by line number somewhere else. It should index by the sign id. Then if you accidentally add two signs on the same line you'll have two signs and know both id's instead of two signs and only know one id.

autozimu:
Thanks!

jimmycuadra:
Was this supposed to handle both warnings and errors? After pulling down the latest commits it seems to have resolved the problem for the latter but not the former.

everett1992:
I'd expect this to fix both warnings and errors

everett1992:
I recommend following my debug process, checking the sign list with `:sign
place` and checking the log.

https://github.com/autozimu/LanguageClient-neovim/issues/952#issuecomment-575420311

PR #962, opened 2020-01-27
Refactor 6d85c80, prevent stale signs

everett1992:
My previous commit fixed an issue that caused stale signs.

The root cause is `state.signs` was a map of line number to sign, and
some faulty code that allowed a set of signs that included two signs
with for the same line to be recorded in vim. State would track the
final sign for a line and loose the reference to the previous signs.
Without a reference the signs cannot be removed.

My previous fix addressed the second level bug, two signs for the same
line in the set of next signs, by sorting the diagnostics by line number
before using `group_by`. Sorting is compute (n log(n)) and memory
(collects the entire iterator into a vec) expensive so this commit
replaces `sorted_by_key` and `group_by` with a loop that accumulates
signs into a map.

This commit also fixes the root cause by changing `state.signs` to a map
by sign id instead of by line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,11 +185,6 @@ version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "either"
-version = "1.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "failure"
 version = "0.1.5"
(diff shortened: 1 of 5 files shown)


martskins:
This is stale and it should no longer be an issue so I'm closing it, but do re-open if you feel like there's still something to address and that this can help.

Thank you for contributing in any case! 🙇

PR #965, opened 2020-01-28
Add support for `rust-anaylzer.selectAndApplySourceChange`

### LINKED ISSUE #963: Support for auto import
I wanted to try the [new `auto import` feature from rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/pull/2887), but unfortunately it does not seem to work out of the box for LanguageClient.

I don't know if it is something that could be implemented easily or if it requires non trivial efforts. In any case I think it could be something interesting for a Rust codebase :wink:

martskins:
This PR fixes #963 by adding support for `rust-analyzer`'s `selectAndApplySourceChange` command, which is needed for the newly introduced `auto import` feature.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -847,6 +847,23 @@ impl LanguageClient {
                     }
                 }
             }
+            "rust-analyzer.selectAndApplySourceChange" => {
+                if let Some(ref edits) = cmd.arguments {
+                    for edit in edits {
+                        let workspace_edits: Vec<WorkspaceEditWithCursor> =
+                            serde_json::from_value(edit.clone())?;
+                        for edit in workspace_edits {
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "languageclient"
-version = "0.1.156"
```
autozimu: Changes here is not needed.

They will be automatically updated with other places as well when cut new release.

autozimu (changes requested):
LGTM. Just small changes to clean up.

Cargo.toml
martskins: Dropped that change on install.sh as well.

autozimu:
Yes please

martskins:
Done!

autozimu:
Thanks!

PR #968, opened 2020-02-02
[WIP] Less clones, less allocations

dodomorandi:
Added an utility module to help `serde_json::Value` deserialization
using a borrow instead of an owned value.

The documentation of the new `json_utils` module is still missing. Feel free to give me some feedback, in order to have better names or similar.

P.S.: I really liked the idea of creating an extension trait for `serde_json::Value`, but with the actual limitations of not being able to return an `impl Trait`, it was kinda impossible or anti-ergonomic to return complex iterator types.

### DIFF
--- a/src/json_utils.rs
+++ b/src/json_utils.rs
@@ -1,0 +1,115 @@
+use failure::Fail;
+use serde::de::DeserializeOwned;
+use serde_json::Value;
+
+pub(crate) fn from_value<T>(value: &Value) -> serde_json::Result<T>
+where
+    T: DeserializeOwned,
+{
+    T::deserialize(value)
+}
+
+pub(crate) fn from_value_opt<T>(value: &Value) -> serde_json::Result<Option<T>>
+where
+    T: DeserializeOwned,
+{
+    match value {
+        Value::Null => Ok(None),
+        _ => T::deserialize(value).map(|value| Some(value)),
+    }
+}
+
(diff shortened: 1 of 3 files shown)


martskins:
Just my two cents on this: I would argue that this adds a bit of complexity to the code base and I'm not sure if it's worth it.
I do however think that we could just replace most of the calls to `serde_json::from_value` that have the clones with `T::deserialize`, avoiding those clones and keeping the simple approach as well.

dodomorandi:
I partially agree with you: I totally don't know the real gain of these changes, and obviously they are adding some more complexity of the code. On the other hand, most of the time using just `T::deserialize` does not help so much. In fact the only interesting gains (if there are any measurable) are related to `Value::Array(Vec<T>)` and similar, where an entire `Vec` of `Value`s is cloned in order to create a `Vec<T>` just to (most of the time) iterate over `T`s.

TL;DR: maybe it could be not worth merging my PR, and in that case it could be reasonable to continue using `serde_json::from_value`.

martskins:
@dodomorandi as of today, the dev branch uses `T::deserialize(&D)` to deserialize most of the params and results, so that effectively removed the need for those extra `.clone()` calls. So I think we can say that this is now a solved issue?  Based on that, I'm gonna close this, but feel free to re-open if you consider that something else should/could be done.

PR #972, opened 2020-02-10
Add config value for preferred markup kind

### LINKED ISSUE #836: MarkupKind is not properly set
This language client didn't set its `MarkupKind` in the `TextDocumentClientCapabilities`, therefore language servers may just pick one from their own preferences. 

1. AFAIK, most vim client doesn't have the real GUI ability to render markdown, sometimes it can be worse than `plaintext`, since `plaintext` are usually optimized for TUI.

2. Noticed that even we can set `2.21 g:LanguageClient_hoverPreview` to `Never`, language servers are likely to response more contents when `markdown` is assumed thus won't fit in one line.

Since `plaintext` also has its own drawbacks, we could introduce a `g` variable for plugin users to choose, ideally per language server.

Or, we could set the `MarkupKind` to `plaintext` when `hoverPreview`  is `Never`. But I think one might want some combinations such as `FloatingWindow + Plaintext` for simplicity.

This is softly connected to #835 which downgraded the preview window experience. Hopefully neovim's floating window will soon landed.

martskins:
This PR fixes #836 by setting `MarkupKind` to `MarkupKind::PlainText` everywhere that's possible. The current implementation results in the server choosing which format to use, and can result, for example, in this:

Screenshot from 2020-02-10 20-30-32

(Note that the documentation for the symbol is enclosed by backticks and includes the language as well.)

This PR will indicate the server that it prefers plain text over markup so it will result (in servers that choose to honor the `MarkupKind` sent by the client) in this:

Screenshot from 2020-02-10 20-30-56

An improvement over this could be to make this a setting, and have the user choose it's `MarkupKind` for each server (or as client-level option if we wanted something simpler), but given that markup support in vim is rather poor, I think this would fit most of the users.

**EDIT**:

Updated this. Now it includes a config variable LanguageClient_preferredMarkupKind which is an array of strings which can be plaintext or markdown. The default value is v:null, which will let the server decide what to use as a markup kind.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -125,6 +125,7 @@ impl LanguageClient {
             semanticHighlightMaps,
             semanticScopeSeparator,
             applyCompletionAdditionalTextEdits,
+            preferred_markup_kind,
         ): (
             Option<u64>,
             String,
@@ -135,6 +136,7 @@ impl LanguageClient {
             HashMap<String, HashMap<String, String>>,
             String,
             u8,
+            Option<Vec<String>>,
         ) = self.vim()?.eval(
             [
                 "get(g:, 'LanguageClient_diagnosticsSignsMax', v:null)",
@@ -146,6 +148,7 @@ impl LanguageClient {
                 "s:GetVar('LanguageClient_semanticHighlightMaps', {})",
                 "s:GetVar('LanguageClient_semanticScopeSeparator', ':')",
                 "get(g:, 'LanguageClient_applyCompletionAdditionalTextEdits', 1)",
+                "get(g:, 'LanguageClient_preferredMarkupKind', v:null)",
             ]
             .as_ref(),
         )?;
@@ -223,6 +226,18 @@ impl LanguageClient {
         let semanticHlUpdateLanguageIds: Vec<String> =
             semanticHighlightMaps.keys().cloned().collect();
 
+        let preferred_markup_kind: Option<Vec<MarkupKind>> = if preferred_markup_kind.is_some() {
+            serde_json::from_value(Value::Array(
+                preferred_markup_kind
+                    .unwrap()
+                    .into_iter()
+                    .map(|x| Value::String(x))
+                    .collect(),
+            ))?
+        } else {
+            None
+        };
+
         self.update(|state| {
             state.autoStart = autoStart;
             state.semanticHighlightMaps = semanticHighlightMaps;
@@ -258,6 +273,7 @@ impl LanguageClient {
             state.loggingLevel = loggingLevel;
             state.serverStderr = serverStderr;
             state.is_nvim = is_nvim;
+            state.preferred_markup_kind = preferred_markup_kind;
             Ok(())
         })?;
 
@@ -1119,6 +1135,7 @@ impl LanguageClient {
         };
 
         let trace = self.get(|state| state.trace)?;
+        let preferred_markup_kind = self.get(|state| state.preferred_markup_kind.clone())?;
 
         let result: Value = self.get_client(&Some(languageId.clone()))?.call(
             lsp::request::Initialize::METHOD,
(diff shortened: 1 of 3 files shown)


theHamsta:
Why? I get nice syntax highlighting of embedded code blocks with markdown. You can always use the conceal feature to hide the backticks.

DerWeh:
I agree with @theHamsta, the default behavior should be getting markup. Maybe we should add documentation for some “recommended standard” plug-in to nicely display the documentation?

Or is it even an alternative to ship the language client with some default highlighting+concealing and set a specific document type? (Of course only if this isn't too complex.)

martskins:
Ah you're right, it seems like I was just trying the wrong markdown plugin.
In any case, I still believe that this should be configurable, I for one would like to have plain text there, and it's a parameter to be sent to the server after all.
I'll update this PR to introduce a config variable for this and send that instead of just hard-coding it to PlainText.

mkatychev:
@DerWeh @theHamsta  do you friends have any suggestions on setup? 

I have this in `ftplugin/markdown.vim`:
```vim
let g:indentLine_enabled = 0
set conceallevel=0
autocmd BufReadPre __LanguageClient__ set conceallevel=3
```

1. This however doesn't appropriately conceal markdown when the floating window is brought up until I place a cursor inside of it. 

1. Syntax highlighting for fenced markdown will only kick in a couple of seconds after the cursor is placed inside.

The syntax highlighting delay only seems to be present in non file buffers and floating windows, real files will pull down the appropriate hi tags in a timely manner. (I've tested this with only the `autozimu/LanguageClient-neovim`, `tpope/vim-markdown` || `plasticboy/vim-markdown` plugins turned on).

theHamsta:
I'm using `tope/vim-markdown` with conceallevel=2. I've encountered the "text not visible until you enter with cursor" issue in go. However, I'm using Python/Rust most of the time so this wasn't a problem for me.

Hope highlighting get's better with tree-sitter. Maybe, it would still be good for people who prefer plain text over markdown to choose. For me personally, I still prefer syntax highlighting of code blocks (has to be activated in tope/markdown) even if conceal is not enabled.

mkatychev:
@theHamsta  are you encountering any delay for syntax highlighting with floating windows in particular?

theHamsta:
@mkatychev no, not at all.

martskins:
Updated this. Now it includes a config variable `LanguageClient_preferredMarkupKind` which is an array of strings which can be `plaintext` or `markdown`. The default value is `v:null`, which will let the server decide what to use as a markup kind.

src/language_server_protocol.rs
```diff
@@ -146,9 +148,11 @@ impl LanguageClient {
                 "s:GetVar('LanguageClient_semanticHighlightMaps', {})",
                 "s:GetVar('LanguageClient_semanticScopeSeparator', ':')",
                 "get(g:, 'LanguageClient_applyCompletionAdditionalTextEdits', 1)",
+                "get(g:, 'LanguageClient_preferredMarkupKind', v:null)",
             ]
             .as_ref(),
         )?;
+        error!("{:?}", preferred_markup_kind);
```
autozimu: debug?

src/language_server_protocol.rs
```diff
@@ -258,6 +262,14 @@ impl LanguageClient {
             state.loggingLevel = loggingLevel;
             state.serverStderr = serverStderr;
             state.is_nvim = is_nvim;
+            if let Some(preferred_markup_kind) = preferred_markup_kind {
```
autozimu: Let's move this block outside `update`. For one, update closure should be fast to avoid locking global state for a long time. Secondly, if error occurred during deserialization, it would avoid the state partially synced.

autozimu (changes requested):
Thanks for contributing!

Two small comments.

src/language_server_protocol.rs
martskins: Oops! :sweat_smile: Solved now!

autozimu:
Thanks!

PR #973, opened 2020-02-11
make command arguments optional

euclio:
It's not currently allowed to send a command that has no arguments.


autozimu:
Thanks for contributing!

PR #976, opened 2020-02-24
Add default root path implementation for Go

bouk:
When inside a go mod, the root of the project is wherever the go.mod file is. Right now the language server fails to initialize if the module is in a subfolder of your source repository.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -63,6 +63,7 @@ pub fn get_rootPath<'a>(
         "scala" => traverse_up(path, |dir| dir.join("build.sbt").exists()),
         "haskell" => traverse_up(path, |dir| dir.join("stack.yaml").exists())
             .or_else(|_| traverse_up(path, |dir| dir.join(".cabal").exists())),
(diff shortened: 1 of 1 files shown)


martskins:
How does this affect users not using go modules?

bouk:
If a user has GO111MODULE=off explicitly but they do have go.mod in a parents directory, then this would do the wrong thing. We could look at the result of running `go env GOMOD` alternatively. Or possibly look at the GO111MODULE environment variable

martskins:
I was referring more specifically to the users of `dep` (and the other dependency management tools around).
I would expect this to attempt to try and find the `go.mod` file and if that fails just go back to the default behavior (using the `rootPaths` config).

bouk:
Yeah but that would be fine. The thing with modules is that it can change how packages are resolved, so I think that's why it can break. If you're not using go modules and just putting your packages in GOPATH, then gopls will be able to find them.

autozimu:
Thanks for contributing!

PR #978, opened 2020-02-27
Do not apply additional text edits if completionPreferTextEdit is set to 0

### LINKED ISSUE #977: Have `completionPreferTextEdit` toggle the handling of `additional text edit`s.
This is more of a question than a feature request or a bug report. Currently, `additional text edits` from a `completionItem/resolve` call are being processed even though the `LanguageClient_completionPreferTextEdit` option is set to `0`. I think it would make sense for those text edits not to happen if the user has the option set to `0`.
Another alternative would be to add an option specifically for this, but I think it makes sense to handle that under the same flag.

So the question is, should we move [this line](https://github.com/autozimu/LanguageClient-neovim/blob/next/src/language_server_protocol.rs#L3101) inside the `if block` where we check the value of `LanguageClient_completionPreferTextEdit`?

This would be a fairly straight forward change, so if this is something that we want to see I can open a PR for it.

martskins:
This PR adds a `LanguageClient_applyCompletionAdditionalTextEdits` flag which can take a value of `0` or `1`. If `1`, it will apply additional text edits for completion items on `completionItem/resolve`, otherwise it won't apply them.

The flag's default value is `1`.

Closes #977.

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -500,6 +500,14 @@ Example configuration for eclipse.jdt.ls:
      highlight! JavaStaticMemberFunction ctermfg=Green cterm=none guifg=Green gui=none
      highlight! JavaMemberVariable ctermfg=White cterm=italic guifg=White gui=italic
 
+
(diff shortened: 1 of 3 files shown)


autozimu:
Awesome! Thanks

PR #996, opened 2020-03-24
Use FZF to display code lens actions

martskins:
This PR uses FZF to display code lens actions. Previously, if multiple code lens actions existed at one line it would run the first one, ignoring the others.

This also adds the implementation of `rust-analyzer.showReferences` action and fixes the display logic for multiple code lenses in the same line.

Peek 2020-03-24 20-32


autozimu:
Thanks!

PR #997, opened 2020-03-25
Hide errors with ContentModified error code

martskins:
This PR ignores the errors with code `-32801` as suggested by the specification:

> if a server detects a state change that invalidates the result of a request in execution the server can error these requests with ContentModified. If clients receive a ContentModified error, it generally should not show it in the UI for the end-user. Clients can resend the request if appropriate.

Not ignoring this error causes the server to display an error to the user and fail to start under some circumstances and with certain servers (I have seen it mostly with `rust-analyzer`).

Sorry about not dropping a link to the doc, but the section in which that is (`Implementation considerations`) is not easily linkable.

### DIFF
--- a/src/rpcclient.rs
+++ b/src/rpcclient.rs
@@ -82,24 +82,7 @@ impl RpcClient {
         // TODO: duration from config.
         match rx.recv_timeout(Duration::from_secs(60))? {
             rpc::Output::Success(ok) => Ok(serde_json::from_value(ok.result)?),
-            rpc::Output::Failure(err) => {
-                // NOTE: Errors with code -32801 correspond to the protocol's ContentModified error,
-                // which we don't want to show to the user and should ignore, as the result of the
-                // request that triggered this error has been invalidated by changes to the state
-                // of the server.
-                if err.error.code.code() == -32801 {
-                    let val = serde_json::from_value(Value::Null);
-                    match val {
-                        Ok(val) => {
-                            return Ok(val);
-                        }
-                        // do not return deserialization errors, as to preserve the original error.
-                        Err(e) => error!("Failed to deserialize null into value: {} ", e),
-                    }
-                }
-
-                bail!("Error: {:?}", err)
-            }
+            rpc::Output::Failure(err) => bail!("Error: {:?}", err),
         }
     }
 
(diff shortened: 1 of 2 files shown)


martskins:
Giving this a second look, I'm not sure if trying to deserialize from `Value::Null` is actually a good idea, as non `Option<T>`s will probably fail, but I can't think of a way around this that doesn't involve modifying the signature of `call`.

src/rpcclient.rs
```diff
@@ -82,7 +82,24 @@ impl RpcClient {
...
         match rx.recv_timeout(Duration::from_secs(60))? {
             rpc::Output::Success(ok) => Ok(serde_json::from_value(ok.result)?),
-            rpc::Output::Failure(err) => bail!("Error: {:?}", err),
+            rpc::Output::Failure(err) => {
+                // NOTE: Errors with code -32801 correspond to the protocol's ContentModified error,
+                // which we don't want to show to the user and should ignore, as the result of the
+                // request that triggered this error has been invalidated by changes to the state
+                // of the server.
+                if err.error.code.code() == -32801 {
+                    let val = serde_json::from_value(Value::Null);
```
autozimu: Rather than substituting errors with nulls and pretending we're getting values back, can we move the logic to https://github.com/autozimu/LanguageClient-neovim/blob/next/src/rpchandler.rs#L12 ?

martskins: Haven't been able to reproduce reliably as I'm not sure how to trigger that error from the server, but I think this should be done now, moved the error handling into handle_call as you suggested.

**EDIT**: No, this doesn't work, hold this off please, I'll try and find another way around.

autozimu:
Thanks!

martskins:
Hm, seems like you missed my edit in my last comment :smile: . This approach doesn't work as the error we get in rpcclient is obfuscated by the call to `bail!`, which returns a `failure::ErrorMessage`, which will never match `jsonrpc_core` error.

I don't think having it there will cause any issues, but if you want to revert this that's ok, I can open a new PR later on.

martskins:
I just found a way around this. If you revert this PR I'll reopen with the correct solution, or I can just open a new one on top of this one to fix this. Let me know what you think is best.

autozimu:
Either is fine with me. Would be great if you could submit PR on top of
current HEAD.

PR #998, opened 2020-03-25
Use AtomicU64 for RpcClient::id counter

martskins:
This PR changes the type of the `id` field in `RpcClient` from `Arc<Mutex<Id>>` to `AtomicU64` and moves the `Arc` to the `State`.


autozimu:
Thanks!

PR #999, opened 2020-03-26
Clean up codeLens code

martskins:
This PR cleans up the code around CodeLens and virtual texts, moving blocks of code to separate functions.

This also fixes an issue where code lenses would not be drawn until the cursor moved.

This will also serve as the base for another PR around CodeLens and parallelizing the calls to `codeLens/resolve` that I have ready to open.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -2067,38 +2067,29 @@ impl LanguageClient {
 
                 let results: Value = client.call(lsp::request::CodeLensRequest::METHOD, &input)?;
                 let code_lens: Option<Vec<CodeLens>> = serde_json::from_value(results.clone())?;
+                let mut code_lens: Vec<CodeLens> = code_lens.unwrap_or_default();
 
-                if code_lens_provider.resolve_provider.is_some() {
-                    let mut resolved_code_lens = vec![];
-                    if let Some(code_lens) = code_lens {
-                        for item in code_lens {
-                            let mut item = item;
-                            if let Some(_d) = &item.data {
-                                if let Some(cl) =
-                                    client.call(lsp::request::CodeLensResolve::METHOD, &item)?
-                                {
-                                    item = cl;
-                                }
+                if code_lens_provider.resolve_provider.unwrap_or_default() {
+                    code_lens = code_lens
+                        .into_iter()
+                        .map(|cl| {
+                            if cl.data.is_none() {
+                                return cl;
                             }
-                            resolved_code_lens.push(item);
-                        }
-                    }
 
-                    self.update(|state| {
-                        state
-                            .code_lens
-                            .insert(filename.to_owned(), resolved_code_lens);
-                        Ok(Value::Null)
-                    })?;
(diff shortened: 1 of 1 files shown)


src/language_server_protocol.rs
```diff
@@ -3123,10 +3096,66 @@ impl LanguageClient {
...
+                let text = command.title;
+
+                match virtual_texts
+                    .iter()
+                    .position(|v: &VirtualText| v.line == line)
+                {
+                    Some(idx) => virtual_texts[idx]
+                        .text
+                        .push_str(format!(" | {}", text).as_str()),
+                    None => virtual_texts.push(VirtualText {
```
autozimu, line 3147: Have you considered `impl From<CodeLens> for VirtualText`?

src/language_server_protocol.rs
```diff
@@ -3123,10 +3096,66 @@ impl LanguageClient {
...
+    ) -> Fallible<Vec<VirtualText>> {
+        let mut virtual_texts = vec![];
+        let diagnostics = self.get(|state| state.diagnostics.clone())?;
+        let diagnosticsDisplay = self.get(|state| state.diagnosticsDisplay.clone())?;
+        let diag_list = diagnostics.get(filename);
+        if let Some(diag_list) = diag_list {
+            for diag in diag_list {
+                if viewport.overlaps(diag.range) {
+                    virtual_texts.push(VirtualText {
+                        line: diag.range.start.line,
```
autozimu, line 3115: Similarly, `impl From<Diagnostic> for VirtualText`

src/language_server_protocol.rs
```diff
@@ -3060,56 +3054,35 @@ impl LanguageClient {
                 .notify("s:AddHighlights", json!([source, highlights]))?;
         }
 
+        let bufnr = self.vim()?.get_bufnr(&filename, params)?;
+        self.draw_virtual_texts(&filename, viewport, bufnr)?;
```
autozimu, line 3058: I feel this is getting more imperative than I'd imagine.

Ideally, I plan to move states into a standalone struct, such that when diagnostics, codelens etc, it internally update vim display to keep sync. No need to do in this PR though. Just for sharing ideas.

src/language_server_protocol.rs
martskins: Yeah but it isn't that handy because the virtual text should, at least in the current implementation, be built only if the code lens has a `command` that is not `None`. And also the the text of the final code lens depends on whether other virtual texts are in place in that line, that's why I opted for a standalone function instead.

src/language_server_protocol.rs
martskins: This one is easier than `impl From<CodeLens> for VirtualText` but still, diagnostics virtual texts have their `hl_group` that is obtained from the config of the language server. So the implementation of `From` isn't really possible. Something similar could be possible but it wouldn't be that different than as it is now.

src/language_server_protocol.rs
autozimu: Make sense. Will merge this in.

PR #1000, opened 2020-03-27
Add `async/await` and use it for codeLens resolution

martskins:
This PR adds `tokio` and `futures` to the dependency list to make use of `async/await` in the codeLens resolution logic, to parallelize the requests to the server.

This PR is based on #999, so the diff for this one should be smaller after we merge that one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,3 +36,5 @@ glob = "0"
 notify = "4"
 shellexpand = "1"
 derivative = "2"
+tokio = { version = "0.2.13", features = [ "rt-threaded", "macros", "blocking" ] }
+futures = { version = "0.3.4" }
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -7,6 +7,7 @@ use crate::lsp::request::Request;
 use crate::rpcclient::RpcClient;
 use crate::sign::Sign;
 use failure::err_msg;
+use futures::executor::block_on;
 use itertools::Itertools;
 use notify::Watcher;
 use std::sync::mpsc;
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -35,3 +35,5 @@ regex = "1"
 glob = "0"
 notify = "4"
 shellexpand = "1"
+tokio = { version = "0.2.13", features = [ "full" ] }
```
autozimu: I'm afraid this might pull in lots lots of crates.

src/language_server_protocol.rs
```diff
@@ -23,14 +25,15 @@ impl LanguageClient {
...
+    pub async fn loop_call(&self, rx: &crossbeam::channel::Receiver<Call>) -> Fallible<()> {
         for call in rx.iter() {
             let language_client = LanguageClient {
                 version: self.version.clone(),
                 state_mutex: self.state_mutex.clone(),
                 clients_mutex: self.clients_mutex.clone(), // not sure if useful to clone this
             };
-            thread::spawn(move || {
+
+            tokio::task::spawn_blocking(move || {
```
autozimu: From what I'm reading of the doc, this will block the caller thread, i.e., it won't process the next message until the previous one is completed.

Whereas here what we want is concurrently processing all the requests.

autozimu (commented):
Thanks. Please check comments inline.

Cargo.toml
martskins: Fixed!

src/language_server_protocol.rs
martskins: Ah yeah you're right. I was missing an `.await` there and I guess that I was also lucky that it didn't block but yeah, `tokio::spawn` is what we want there instead. Fixed!

Cargo.toml
```diff
@@ -35,3 +35,5 @@ regex = "1"
 glob = "0"
 notify = "4"
 shellexpand = "1"
+tokio = { version = "0.2.13", features = [ "rt-core", "rt-threaded", "macros", "blocking" ] }
+futures = { version = "0.3.4", features = [ "thread-pool" ] }
```
autozimu: Why thread-pool feature is needed here?

Cargo.toml
```diff
@@ -35,3 +35,5 @@ regex = "1"
 glob = "0"
 notify = "4"
 shellexpand = "1"
+tokio = { version = "0.2.13", features = [ "rt-core", "rt-threaded", "macros", "blocking" ] }
```
autozimu: I'm not familiar with tokio, but do all the features are required here?

martskins: Actually, I think we can ditch `rt-core`, as we have `rt-threaded` already. The other are two needed. Updated to remove that extra feature.

Cargo.toml
martskins: It isn't. I used `FuturesUnordered` before the first changes made to the PR and then it was needed, but it isn't anymore. Removed it!

autozimu:
FYI, this PR cannot be merged as of now due to conflicts.

martskins:
Rebased and solved the conflicts!

martskins:
Closing this, as there's really not much to gain from it and introduces unneeded complexity.

Issue #993, opened 2020-03-17
Add ability to check language server status per filetype


nkakouros:
## Is your feature request related to a problem? Please describe.

I want to have an icon in my status line when a language server is running
(not just idle/busy) for the file that I am currently editing.

## Describe the solution you'd like

A function or a variable that I can use to check if a language server for the
current filetype is running. Sth like `LanguageClient#status('python')`.

## Describe alternatives you've considered

I tried using what is already there but I couldn't make it work. Do you have
any tips to share?

martskins:
This is already possible via the `b:LanguageClient_isServerRunning` variable. See #1003

Issue #742, opened 2019-02-25
Statusline integration


ghost:
## Integrating into Statusline 
LanguageClient-neovim is great, it's extremely fast and I love that it is built on rust.
I use [Ale](https://github.com/w0rp/ale) for linting languages that don't have stable language servers, I've noticed their README has examples on how to integrate Ale into [Lightline](https://github.com/itchyny/lightline.vim)(My favorite Statusline) and for vim's builtin statusline(Hardecore Vimmer's Statusline)

## Solution
I think [Ale](https://github.com/w0rp/ale) has a very well filled out README and I'd like to see LanguageClient-neovim's README at the same level, but I'd especially like to see how to integrate it into my lightline status line without using a [plugin](https://github.com/Palpatineli/lightline-lsc-nvim) and into a normal status line so that people looking to make the new Airline | Lightline | Powerline alternative can have day 1 LanguageClient-neovim integration support

jubnzv:
Another solution is to write [a small function](https://github.com/jubnzv/dotfiles/blob/master/vim/.config/nvim/init.vim#L773) in your `.vimrc` to display LanguageClient status:
```vim
augroup LanguageClient_config
  au!
  au User LanguageClientStarted call LSPUpdateStatus(1)
  au User LanguageClientStopped call LSPUpdateStatus(0)
augroup END
let g:lsp_status = 0
function! LSPUpdateStatus(status) abort
  let g:lsp_status = a:status
  call lightline#update()
endfunction
function! LightlineLSPStatus() abort
  return g:lsp_status == 1 ? 'LSP' : ''
endfunction
```

And add it to [lightline configuration](https://github.com/jubnzv/dotfiles/blob/master/vim/.config/nvim/init.vim#L326) like this:
```vim
let g:lightline = {
  \ 'component_expand': {
  \   'lsp_status': 'LightlineLSPStatus'
  \ },
  \ }
```

Kotaro7750:
jubnz's way does not work when you edit two files including file which lsp is not working for. 
For example, assuming that you enabled cpp-languageserver, when you open cpp file and open txt file in another buffer, g:lsp_status remains 1 in txt file buffer.

Kotaro7750:
Although I have not checked source code yet, a function like LanguageClient#isServerRunningForCurrentBuffer will solve this problem.

```vim
function! LightlineLSPStatus() abort
  return LanguageClient#isServerRunningForCurrentBuffer() ? 'LSP' : ''
[...]

PR #1003, opened 2020-04-04
new function LanguageClient#isServerRunning() for statusline integration

Kotaro7750:
As written in issue #993 #742, many users including me want the feature that indicates whether language server is running for current buffer or not in purpose of displaying status in their statusline.

A new function 'LanguageClient#isServerRunning()' which I implemented enables users to display LSP status to their statusline.

For example in lightline.vim,
```vim
function! LangugeServerStatus() abort
  let l:errors = 0
  let l:warnings = 0
  for item in getqflist()
      if item["type"] == "E"
          let l:errors += 1
      else 
        if item["type"] == "W"
          let l:warnings += 1
        endif
      endif
  endfor
  if LanguageClient#isServerRunning() == 0
    return 'not working!'
  endif
  return l:errors + l:warnings == 0 ? "✔ " : "✖ :" . l:errors . " " . "⚠ :" . l:warnings
endfunction

```
.vim file, no language server

Screenshot from 2020-04-04 14-28-28

.go file, gopls is working!
Screenshot from 2020-04-04 14-29-50

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1050,6 +1050,7 @@ function! LanguageClient#handleBufNewFile() abort
 endfunction
 
 function! LanguageClient#handleBufEnter() abort
+    let b:LanguageClient_isServerRunning = 0
     try
         call LanguageClient#Notify('languageClient/handleBufEnter', {
                     \ 'filename': LSP#filename(),
@@ -1299,11 +1300,8 @@ function! LanguageClient#serverStatusMessage() abort
     return g:LanguageClient_serverStatusMessage
 endfunction
 
-" Set to 1 when the language server is running for current buffer
-let g:LanguageClient_isServerRunning = 0
-
 function! LanguageClient#isServerRunning() abort
-    return g:LanguageClient_isServerRunning
+    return b:LanguageClient_isServerRunning
 endfunction
 
 " Example function usable for status line.
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1289,6 +1289,13 @@ function! LanguageClient#serverStatusMessage() abort
     return g:LanguageClient_serverStatusMessage
 endfunction
 
+" Set to 1 when the language server is running for current buffer
+let g:LanguageClient_isServerRunning = 0
+
+function! LanguageClient#isServerRunning() abort
+    return g:LanguageClient_isServerRunning
+endfunction
+
 " Example function usable for status line.
 function! LanguageClient#statusLine() abort
     if g:LanguageClient_serverStatusMessage ==# ''
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1049,6 +1049,16 @@ function! LanguageClient#handleBufNewFile() abort
     endtry
 endfunction
 
+function! LanguageClient#handleBufEnter() abort
+    try
+        call LanguageClient#Notify('languageClient/handleBufEnter', {
+                    \ 'filename': LSP#filename(),
+                    \ })
+    catch
+        call s:Debug('LanguageClient caught exception: ' . string(v:exception))
+    endtry
+endfunction
+
 function! LanguageClient#handleFileType() abort
     try
         if s:Debounce(2, 'LanguageClient#handleFileType')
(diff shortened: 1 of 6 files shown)


Kotaro7750:
In previous commit 75071f2 , when you open fileA(lsp-enabled) and fileB(lsp-unable) in this order, then re-open fileA,statusline of fileA's buffer says 'not working!'.
This is because languageClient_handleFileType() is not called when you open existing buffer.
So, I stopped changing function languageClient_handleFileType().
To solve this problem, I must create new function languageClient_handleBufEnter().

In this commit, I created languageClient_handleBuffer() to make it possible to change `g:LanguageClient_isServerRunning` when enter existing buffer.
But by only this change, when you open fileA(lsp-enabled filetype),  `g:LanguageClient_isServerRunning` is set to 0 because language server does not start before bufenter.
So, I sets `g:LanguageClient_isServerRunning` to 1 in languageClient_startServer() and also set it to 0 in cleanup().

@autozimu Could you review my code?

autozimu:
Are you aware of this similar function [`languageClient_isAlive`](https://github.com/autozimu/LanguageClient-neovim/blob/1ff138df7ec8a8c35f21eadf8f8d3df53300ec21/src/language_server_protocol.rs#L2687-L2694)?

There is also [`LanguageClientStarted`](https://github.com/autozimu/LanguageClient-neovim/blob/1ff138df7ec8a8c35f21eadf8f8d3df53300ec21/src/language_server_protocol.rs#L3774) and [`LanguageClientStopped`](https://github.com/autozimu/LanguageClient-neovim/blob/1ff138df7ec8a8c35f21eadf8f8d3df53300ec21/src/language_server_protocol.rs#L1080) autocmds.

Kotaro7750:
@autozimu 
Yes, I am aware of `languageClient_isAlive`, but this function is not useful a little in terms of using in statusline because  we must set callback function to receive response.
In statusline function, a function which simply returns boolean whether language server is running is desired, so I created `languageClient_isServerRunning`
This function returns `g:LanguageClient_IsServerRunning` which is set  when we switch buffer, language server starts and stops, so we don't have to set callback.

Also, those two event fired only when language server start or stop, not when we change buffer, so those are not useful for statusline too.

autoload/LanguageClient.vim
```diff
@@ -1289,6 +1299,13 @@ function! LanguageClient#serverStatusMessage() abort
     return g:LanguageClient_serverStatusMessage
 endfunction
 
+" Set to 1 when the language server is running for current buffer
+let g:LanguageClient_isServerRunning = 0
```
autozimu: This is a global variable while we use it as a buffer local one.

I'd suggest we move it to buffer local. Thus we can update it accordingly with various lifecycle autocmds such as BufType, BufNewFile etc.

plugin/LanguageClient.vim
```diff
@@ -98,6 +98,10 @@ function! LanguageClient_serverStatusMessage(...)
     return call('LanguageClient#serverStatusMessage', a:000)
 endfunction
 
+function! LanguageClient_isServerRunning(...)
+    return call('LanguageClient#isServerRunning', a:000)
```
autozimu, line 102: Where is this call being handled?

autozimu (changes requested):
Please check previous comments.

plugin/LanguageClient.vim
Kotaro7750: The function itself is in the autoload directory, and I'm going to handle it there. I did the same for the other functions, but is there any problem?

autoload/LanguageClient.vim
Kotaro7750: I'll change it to buffer local variable and init `b:LanguageClient_isServerRunning` to 0 in `handleBufEnter`, is it ok?

autozimu:
Sounds good

Kotaro7750:
I changed `LanguageClient_isServerRunning` to buffer local variable in commit 23a20e3.

autozimu:
Thanks!

Kotaro7750:
Thank you so much for your thoughtful review.
This is my first OSS contribution, so I'm very happy that my PR is merged.

autozimu:
You're welcome 😄

Issue #1005, opened 2020-04-05
make LanguageClient#statusLine returning diagnostics summary


Kotaro7750:
## Is your feature request related to a problem? Please describe.

`LanguageClient#statusLine()` sounds like returning error count, but returns `g:LanguageClient_serverStatusMessage` which represent rust's build status. I think this behavior is contradicted to function name and this function should return error count like ALE's similar function `ale#statusline#Count`.

## Describe the solution you'd like

I would like to make `LanguageClient#statusLine` returning current buffer's diagnostics in dictionary type or string type.

Ex1. returns "Error:10 Warning:3 Info:2". Users can set directly to statusline message.
Ex2. returns 
```
{
   Error:10,
   Warning:3,
   info:2
}
```
Users can parse this dict and manually set to their statusline.

## Describe alternatives you've considered

For compatibility reasons, this function might as well stay that way. However, I think that functions with features such as  mentioned above would be beneficial to many users, so new function is still needed.

## Additional context

If you'd like, I will implement and make a pull request.

autozimu:
Make sense.

For backwards compatibility, I'd suggest create a new function and deprecate the original one.

Kotaro7750:
Thank you, I'll try it.

PR #1006, opened 2020-04-06
new function LanguageClient#statusLineDiagnosticsCounts

Kotaro7750:
This pull request is a solution of issue #1005 .

As autozimu said in issue #1005, for compatibility, I created new function named `LanguageClient#statusLineDiagnosticsCounts`.

This function returns current buffer's diagnostics dictionary.
Ex. 
```
{'E': 0, 'W': 0, 'H': 0, 'I': 0}
```

By using this function, many users can integrate this very awesome plugin into their status line.

For `main.go`, this function returns 4 errors.
Screenshot from 2020-04-06 21-59-05

For `purchase.go`, returns 2 errors.
Screenshot from 2020-04-06 21-58-59

This function will be more useful if my another pull request #1003 is merged.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -9,7 +9,6 @@ use crate::sign::Sign;
 use failure::err_msg;
 use itertools::Itertools;
 use notify::Watcher;
-use std::collections::BTreeMap;
 use std::sync::mpsc;
 use vim::try_get;
 
@@ -2288,7 +2287,7 @@ impl LanguageClient {
         })?;
         self.update_quickfixlist()?;
 
-        let mut severityCount: BTreeMap<String, u64> = [
+        let mut severityCount: HashMap<String, u64> = [
             (
                 DiagnosticSeverity::Error
                     .to_quickfix_entry_type()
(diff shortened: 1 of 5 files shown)


src/language_server_protocol.rs
```diff
@@ -9,6 +9,7 @@ use crate::sign::Sign;
 use failure::err_msg;
 use itertools::Itertools;
 use notify::Watcher;
+use std::collections::BTreeMap;
```
autozimu: What is the benefit of using BTreeMap here?

Otherwise, LGTM.

Please rebase the PR so I can merge it in.

Kotaro7750: Oh, I first used Map type <DiagnosticsSeverity,u64> and hash does not implemented for DiagnosticsSeverity, and that type is defined in other library, so I must use BTreeMap.
But I use <String,u64> now, so I can use HashMap.
Change and rebase is already done, so I will commit soon.

Kotaro7750:
I changed BTreeMap to HashMap and rebased in 2052b0f

autozimu:
Thanks!

PR #1009, opened 2020-04-10
init b:LanguageClient_isServerRunning only when it is not defined yet.

Kotaro7750:
This is related with my merged pull request #1003 .
In that pull request, I init `b:LanguageClient_isServerRunning` all time BufEnter, so I changed it to be init only when this valiable is not defined yet.


Kotaro7750:
I'm sorry I didn't notice this when I created #1003 .

autozimu:
No problem. Thanks!

PR #1010, opened 2020-04-13
fix:  PathBuf can't be formatted with the default formatter

andoriyu:
PathBuf/Path can't be formatted with `{}` (aka `Display` trait). This causes compilation error.
Solution: Convert PathBuf with PathBuf::display

More information: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.display


andoriyu:
I guess this was already fixed by converting it to string and I didn't notice it.

PR #1013, opened 2020-04-26
Multiple configuration files

### LINKED ISSUE #431: Add support for global project settings
Currently, the file pointed to by `settingsPath` (by default) is a per-project `.vim/settings.json`. However, in many cases, it makes sense to define some settings globally (e.g., `build_on_save`). It'd be nice if LanguageClient read `~/.vim/settings.json` *and* `./.vim/settings.json`, and merged them (with the latter overriding the former).

Second, it'd be fantastic if we could have support for per-language settings similar to [what exists](https://github.com/jonasbb/rls_vscode/blob/e3884bbae7a0012fd7adea9565a2e36343c2f0e5/package.json#L263) in VSCode. For example, if my `settings.json` contains `rust.build_on_save`, the setting `build_on_save` should only be sent if the language is `rust`.

symphorien:
Fixes #431 

With 
```
let g:LanguageClient_settingsPath = [ '~/.config/lsp_global_settings.json', '.vim/settings.json' ]
```
you get a global configuration which can be enriched and overridden by a local one.
The merging is done in the standard way (apparently RFC 7396) via the json-patch crate.


autozimu:
Thanks!

PR #1022, opened 2020-05-10
Add support for codeActions in ranges

### LINKED ISSUE #1016: Code actions for ranges

The code action request supports a range, however currently as far as I can tell LanguageClient-neovim can only send code action requests where the range is set to the current cursor position, and not where the range is set according to the whole current selection.

Code action sends selection range properly.

## Additional context

An example range code action is this introduce variable from rust-analyzer: https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/user/assists.md#introduce_variable

martskins:
This PR adds support for codeActions in visual ranges.

It basically adds a `LanguageClient#textDocument_visualCodeAction` function that uses the current range in the call to the code actions rust handler.

I'm not sure where the best place to document how this should be used is, as the user should use different mappings for each of the code actions functions.

```
vnoremap  ga :call LanguageClient#textDocument_visualCodeAction()<CR>
nnoremap  ga :call LanguageClient#textDocument_codeAction()<CR>
```

Closes #1016

### DIFF
--- a/autoload/LSP.vim
+++ b/autoload/LSP.vim
@@ -52,8 +52,46 @@ function! LSP#viewport() abort
 endfunction
 
 function! LSP#position() abort
+  return s:get_position()
+endfunction
+
+function! LSP#range(mode) abort
+  if a:mode ==# 'v'
+    return s:get_visual_selection()
+  endif
+
+  return {
+    \ 'start': s:get_position(),
+    \ 'end': s:get_position(),
+    \ }
+endfunction
+
+function! s:get_position() abort
 	return {
 		\ 'line': LSP#line(),
 		\ 'character': LSP#character(),
 		\ }
+endfunction
+
+function! s:get_visual_selection() abort
+    let [line_start, column_start] = getpos("'<")[1:2]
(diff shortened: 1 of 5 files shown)


autoload/LSP.vim
```diff
@@ -52,8 +52,45 @@ function! LSP#viewport() abort
...
 		\ 'character': LSP#character(),
 		\ }
 endfunction
+
+function! s:get_visual_selection() abort
+    let [line_start, column_start] = getpos("'<")[1:2]
+    let [line_end, column_end] = getpos("'>")[1:2]
+    let lines = getline(line_start, line_end)
+    if len(lines) == 0
+        return ''
```
autozimu, line 82: Does this cause serialization error when later used in serde?

martskins: I guess it could cause a deserialization error if the user mis-uses this function and ends up with 0 lines there.

I added an error message before the return to signal the user that they are doing something wrong, I can't think of a better way of handling this, as returning a default range is probably misleading, so I guess erroring is the right thing here.

autozimu:
Thanks!

PR #1024, opened 2020-05-10
Silence diagnostic not found error in explainErrorAtPoint

### LINKED ISSUE #884: Can't use LanguageClient#explainErrorAtPoint() in CursorHold autocmd

I have the following autocmd:

```
au CursorHold * sil call LanguageClient#explainErrorAtPoint()
```

It works greater and displays the diagnostic in a floating window with the newest neovim. However, when I'm on a line without error, it outputs the error message "No diagnostic found: ...". This is annoying.

I'd like that calling `explainErrorAtPoint()` on a line without error does nothing.

I've tried rasing the log level, but this doesn't seem to work.

martskins:
This PR silences the error returned by `explainErrorAtPoint` when called on point with no diagnostics in it.

This is useful if you have configured this function to be called on a `CursorHold` autocmd for example.

Closes #884.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -3548,23 +3548,21 @@ impl LanguageClient {
         info!("Begin {}", REQUEST__ExplainErrorAtPoint);
         let filename = self.vim()?.get_filename(params)?;
         let position = self.vim()?.get_position(params)?;
-        let diag = self.get(|state| {
-            state
-                .diagnostics
-                .get(&filename)
-                .ok_or_else(|| format_err!("No diagnostics found: filename: {}", filename,))?
-                .iter()
-                .find(|dn| position >= dn.range.start && position < dn.range.end)
-                .cloned()
-                .ok_or_else(|| {
-                    format_err!(
-                        "No diagnostics found: filename: {}, line: {}, character: {}",
-                        filename,
-                        position.line,
-                        position.character
-                    )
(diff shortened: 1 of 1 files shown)


src/language_server_protocol.rs
```diff
@@ -3548,23 +3548,21 @@ impl LanguageClient {
...
-        })??;
+        let diag = self
+            .get(|state| state.diagnostics.get(&filename).cloned())?
+            .unwrap_or_default()
+            .iter()
+            .find(|dn| position >= dn.range.start && position < dn.range.end)
+            .cloned();
+
+        if diag.is_none() {
+            debug!(
```
autozimu, line 3559: May I suggest adding a user config and default to the old behaviour here?

The reason is because, when user calls this method explicitly, something is expected from user end. If we did nothing here, user might be confused or wondering why the request hasn't fulfilled yet.

Make sense?

martskins: What we could do is add structured logging (on the vim side) and add a config to filter the messages based on the level, as to avoid having a config value for a specific error message, but that sounds like a little more work so maybe we hold that off until we find that it's really needed?

Meanwhile, I'm ok with not merging this if you think it may be confusing for users.

martskins:
Closing this, it would probably better to create a different function that does this so we catch anyone by surprise.

PR #1027, opened 2020-05-11
Implement $/progress notification

martskins:
Adds the implementation for `$/progress`.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -8,6 +8,7 @@ use crate::rpcclient::RpcClient;
 use failure::err_msg;
 use itertools::Itertools;
 use notify::Watcher;
+use serde::de::Deserialize;
 use std::sync::mpsc;
 use vim::try_get;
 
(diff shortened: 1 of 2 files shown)


src/language_server_protocol.rs
```diff
@@ -2041,6 +2042,36 @@ impl LanguageClient {
...
+        }
+
+        let token = match params.token {
+            // number is a not a particularly useful token to report to the user, so we just use
+            // INFO instead.
+            NumberOrString::Number(_) => "INFO".to_string(),
+            NumberOrString::String(s) => s,
+        };
+
+        let message = format!("{}: {}", token, message.unwrap());
```
autozimu: Nit: I'd suggest update `message.unwrap()` => `message.upwrap_or_default()`.

Though message itself has already been checked previously, `None` might slip in in future changes. `unwrap`  would halt the whole program, which is very unfortunate.

I'm actually considering adding a clippy rule to warn/error about `.unwrap`.

PR #1029, opened 2020-05-14
Ignore bin directory

Shatur:
I added bin directory to ignore. This is useful for users who use native package manager. I also sorted ignored list alphabetically.


Shatur:
Sorry, did not notice that bin is already ignored in `next`

PR #1030, opened 2020-05-15
Fix content modified error handling

martskins:
Fixes the error handling around the content modified error coming from the language servers.


autozimu:
Thanks!

PR #1032, opened 2020-05-18
Add missing range argument to visualCodeAction function

martskins:
This PR adds the missing `range` argument to the `visualCodeAction` function.

The effect of this missing was that when selection a range that spans multiple lines, the function would be called once per line in that range, which is not what we want.


autozimu:
Thanks!

Issue #1033, opened 2020-05-22
Use minimal settings for Neovim floating window


mklein994:
- [x] Did you upgrade to latest plugin version?
- [x] Did you upgrade to/compile latest binary? Run shell command
  `bin/languageclient --version` to get its version number.
- [x] \(Neovim users only) Did you check output of `:checkhealth LanguageClient`?
- [x] Did you check [troubleshooting]?

[troubleshooting]: https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#troubleshooting

## Describe the bug

If I have some settings that affect the appearance of the window in my `init.vim` file, they are by default used in Neovim's floating window. This can make it a bit cramped if the screen size is small.

## Environment

- neovim/vim version (`nvim --version` or `vim --version`):
```
NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/lib/distcc/bin/cc -march=armv7-a -mfloat-abi=hard -mfpu=vfpv3-d16 -O2 -pipe -fstack-protector-strong -fno-plt -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.4.3/src -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
- This plugin version (`git rev-parse --short HEAD`):
```
4b35f5d
```
- This plugin's binary version (`bin/languageclient --version`):
```
languageclient 0.1.157 18434f2cd9abc33498e705dfde70d80e19223d0f
```
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue. Refer to an example [here][min-vimrc.vim]):
[...]

PR #1034, opened 2020-05-22
Use a minimal style for Neovim's floating window

mklein994:
Disable many UI options in the floating hover window, such as
'foldcolumn' and 'number'. See `:help nvim_open_win()` for details.

Issue #1033

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -420,6 +420,7 @@ function! s:OpenHoverPreview(bufname, lines, filetype) abort
         \   'col': col,
         \   'width': width,
         \   'height': height,
+        \   'style': 'minimal',
         \ })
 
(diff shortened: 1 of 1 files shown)


mklein994:
My only concern on this is that it should be configurable, but I'm not sure how best to implement that.

autozimu:
LGTM.

User config might be introduced similar to https://github.com/autozimu/LanguageClient-neovim/blob/next/src/language_server_protocol.rs#L109

autozimu:
Let me know if you want to take that up.

Otherwise I can introduce the config support.

mklein994:
I gave it a shot, but I'm in over my head here. I think it's best if you implement it. :slightly_smiling_face:

autozimu:
Thanks!

PR #1039, opened 2020-05-29
Fix incorrect params in textDocument_codeAction call

martskins:
Fixes `LanguageClient_textDocument_codeAction`, as it was passing an extra parameter and causing the function not to work.


autozimu:
Thanks!

PR #1043, opened 2020-06-02
Show completion item documentation on selection change

### LINKED ISSUE #875: Add support for calling LSP completionItem/resolve implementation when completion items are selected
(I originally posted this over in [the deoplete repo](https://github.com/Shougo/deoplete.nvim) ([\#1004](https://github.com/Shougo/deoplete.nvim/issues/1004)) because I thought that might be the right place for it, but @Shougo seems to think this is best handled inside LanguageClient-neovim, so I'm creating this ticket to see if that's correct.)

> Looking at the LSP spec, I see [this](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion) (emphasis added):
>
>>  If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request (‘completionItem/resolve’). This request is sent when a completion item is selected in the user interface. **A typical use case is for example: the ‘textDocument/completion’ request doesn’t fill in the documentation property** for returned completion items since it is expensive to compute. **When the item is selected in the user interface then a ‘completionItem/resolve’ request is sent** with the selected completion item as a parameter. The returned completion item should have the documentation property filled in.
>
> To this effect, LanguageClient-neovim [implemented](https://github.com/autozimu/LanguageClient-neovim/pull/514) a `LanguageClient#completionItem_resolve()` function that can be called to get additional information about a completion. The obvious use-case here is to show info about the selection completing in a preview window (or floating window).
>
> As far as I can tell, LanguageClient-neovim sets its `LanguageClient#complete()` function as the `'omnifunc'`, and deoplete calls that to get completions. But when you move through completions, deoplete doesn't call `LanguageClient#completionItem_resolve()` and so doesn't get the extra info that an LSP server might provide. One such example server would be [typescript-language-server](https://github.com/theia-ide/typescript-language-server), which doesn't provide extended information for `textDocument/completion` requests, but does implement `completionItem/resolve`.
>
> So this is really a feature request. Could we have an option to make deoplete call an LSP `completionItem/resolve` function when completion items are selected? This means that `'completopt=preview'` would be more useful, because it would have information to show about selected completions in more cases. coc.nvim [implements this](https://github.com/neoclide/coc.nvim/blob/3e0eec9c4f08003304ba4d50bb0866dbaf7cc707/src/provider/index.ts#L40-L52), in case you wanted to check out what they're doing.

Looking into `:h CompleteChanged`, I see that support for it was added to Neovim [here](https://github.com/neovim/neovim/pull/9616) by @chemzqm (the [Coc](https://github.com/neoclide/coc.nvim/issues/441) maintainer).

It's still not clear to me exactly where the responsibility for this functionality should lie (that's why I opened the issue in deoplete; but it may be here), but the end goal is to have something like what you can see in the GIF in the Coc README:

https://user-images.githubusercontent.com/251450/55285193-400a9000-53b9-11e9-8cff-ffe4983c5947.gif

You can see how the completion menu appears, and then, asynchronously, a floating preview appears to the side with additional information about the selected completion:

Screen_Shot_2019-08-24_at_21_10_47_png

Thoughts?

martskins:
This is a PR to propose adding a popup window to show completion item documentation when the user walks through the completion menu ~and to kick things off if this is something we wanted to add~.

This uses the `documentation` key in the completion items if provided in the `textDocument/completion` response, or uses `completionItem/resolve` as a fallback to get that documentation. The documentation is displayed reusing the `OpenHoverPreview` function, so it should behave in the same way as that (shows in floating window, popup window or preview, depending on which is configured or available).

~The code is a little messy and there are some things that I'm not completely sure about, but I figured I'd push this and receive feedback and requests for changes anyway.~

This is how it looks in vim:
vim

And in neovim:

nvim

Note that in the vim example, the documentation is available in the `textDocument/completion` response, so no further `completionItem/resolve` calls are made. Whereas in the neovim example, the documentation is not available until we explicitly call `completionItem/resolve`, because the javascript language server does not provide it in the `textDocument/completion` response.

**IMPORTANT**
~This uses `pum_getpos` to get the completion menu position and dimensions, so it won't work in neovim 0.4.X, as it's not available there, but it is available in neovim 0.5 (nightly) and vim 8. We need to add a check for this somewhere.~

**EDIT**
This is now working for nvim 0.4.X and 0.5, as well as vim8.

Closes #875

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -11,6 +11,9 @@ let s:TYPE = {
 let s:FLOAT_WINDOW_AVAILABLE = exists('*nvim_open_win')
 let s:POPUP_WINDOW_AVAILABLE = exists('*popup_atcursor')
 
+" timers to control throttling
+let s:timers = {}
+
 function! s:AddPrefix(message) abort
     return '[LC] ' . a:message
 endfunction
@@ -426,7 +429,9 @@ endfunction
 "   - Floating window on Neovim (0.4.0 or later)
 "   - popup window on vim (8.2 or later)
 "   - Preview window on Neovim (0.3.0 or earlier) or Vim
-function! s:OpenHoverPreview(bufname, lines, filetype, x, y) abort
+"
+" Receives two optional arguments which are the X and Y position 
+function! s:OpenHoverPreview(bufname, lines, filetype, ...) abort
     " Use local variable since parameter is not modifiable
     let lines = a:lines
     let bufnr = bufnr('%')
@@ -493,9 +498,9 @@ function! s:OpenHoverPreview(bufname, lines, filetype, x, y) abort
         endif
 
         let relative = 'cursor'
-        if a:x != v:null && a:y != v:null
-          let col = a:x
-          let row = a:y
+        let col = get(a:000, 0, col)
+        let row = get(a:000, 1, row)
+        if get(a:000, 0, v:null) isnot v:null && get(a:000, 1, v:null) isnot v:null
           let relative = 'win'
         endif
 
@@ -514,10 +519,11 @@ function! s:OpenHoverPreview(bufname, lines, filetype, x, y) abort
         let float_win_highlight = s:GetVar('LanguageClient_floatingHoverHighlight', 'Normal:CursorLine')
         execute printf('setlocal winhl=%s', float_win_highlight)
     elseif display_approach ==# 'popup_win'
-        if a:x != v:null && a:y != v:null
-          let pop_win_id = popup_create(a:lines, { 'line': a:y + 1, 'col': a:x + 1})
+        let l:padding = [1, 1, 1, 1]
+        if get(a:000, 0, v:null) isnot v:null && get(a:000, 1, v:null) isnot v:null
+          let pop_win_id = popup_create(a:lines, { 'line': get(a:000, 1) + 1, 'col': get(a:000, 0) + 1, 'padding': l:padding })
         else
-          let pop_win_id = popup_atcursor(a:lines, {})
+          let pop_win_id = popup_atcursor(a:lines, { 'padding': l:padding })
         endif
         call setbufvar(winbufnr(pop_win_id), '&filetype', a:filetype)
     elseif display_approach ==# 'preview'
@@ -536,7 +542,7 @@ function! s:OpenHoverPreview(bufname, lines, filetype, x, y) abort
 
         call setline(1, lines)
         " trigger refresh on plasticboy/vim-markdown
-        normal! i
+        doautocmd InsertLeave
         setlocal nomodified nomodifiable
 
         wincmd p
@@ -1064,6 +1070,7 @@ function! LanguageClient#completionItem_resolve(completion_item, ...) abort
                 \ 'completionItem': a:completion_item,
                 \ 'handle': s:IsFalse(l:Callback)
                 \ }
+    call extend(l:params, get(a:000, 0, {})) " extend with pumpos params
     return LanguageClient#Call('completionItem/resolve', l:params, l:Callback)
 endfunction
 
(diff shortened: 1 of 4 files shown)


autoload/LanguageClient.vim
```diff
@@ -1502,4 +1512,78 @@ function! LanguageClient#debugInfo(...) abort
...
+
+  if v:completed_item['user_data'] ==# ''
+    call s:ClosePopups()
+    return
+  endif
+
+  let l:user_data = json_decode(v:completed_item['user_data'])
+  let l:completed_item = {}
+
+  if has_key(l:user_data, 'ncm2_lspitem')
```
martskins: I'm not sure why this is needed really, all I know is that the `lspitem` key is `ncm2_lspitem` when running `ncm2`, but I'm not sure if other plugins (deoplete, etc) do this as well or not.

martskins:
Found the way to make this work in nvim `0.4.X`. Will try this out for a few days and merge if everything is working.

autoload/LanguageClient.vim
```diff
@@ -1634,4 +1654,82 @@ function! LanguageClient#debugInfo(...) abort
...
+" receives the v:event from the CompleteChanged autocmd
+function! LanguageClient#handleCompleteChanged(event) abort
+  " this timer is just to stop textlock from locking our changes
+  call timer_start(0, funcref('s:ClosePopups'))
+
+  if exists('b:timers') && has_key(b:timers, 'CompleteChanged')
+    call timer_stop(b:timers['CompleteChanged'])
+  else
+    let b:timers = {}
+  endif
```
martskins, line 1676: This block here is to add throttling to these calls, as displaying the documentation can be a little expensive, especially for markdown documentation.

Also, and unrelated to this, but maybe we want to take this strategy for `CursorMoved` as well? It would save a lot of unnecessary calls to the server.

autoload/LanguageClient.vim
```diff
@@ -1634,4 +1663,81 @@ function! LanguageClient#debugInfo(...) abort
...
+endfunction
+
+" receives the v:event from the CompleteChanged autocmd
+function! LanguageClient#handleCompleteChanged(event) abort
+  " the timers on zero are just to stop textlock from locking our changes
+  if !s:Debounce(100, 'LanguageClient#handleCompleteChanged')
+    call timer_start(0, funcref('s:ClosePopups'))
+    return
+  endif
```
martskins, line 1682: I used `s:Debounce` because it was already being used but I think we should switch to something like `timer_start(100, { -> functionToCall() })` and store the returned timer id in a dictionary keyed by function name and cancel it before calling the same function again. That allows for more precise control, I've been getting very dubious results with this and I suppose it must be either the extra cycles it has to compute to evaluate the conditions for Debounce and the call itself, or maybe calculating the diff with `reltime` just isn't a reliable enough method.

autoload/LanguageClient.vim
```diff
@@ -1634,4 +1657,80 @@ function! LanguageClient#debugInfo(...) abort
...
+" receives the v:event from the CompleteChanged autocmd
+function! LanguageClient#handleCompleteChanged(event) abort
+  " this timer is just to stop textlock from locking our changes
+  call timer_start(0, funcref('s:ClosePopups'))
+
+  if has_key(s:timers, 'LanguageClient#handleCompleteChanged')
+    call timer_stop(s:timers['LanguageClient#handleCompleteChanged'])
+  endif
+
+  let s:timers['LanguageClient#handleCompleteChanged'] = timer_start(100, funcref('LanguageClient#showCompletionItemDocumentation', [a:event]))
```
martskins: I used this instead of the existing `s:Debounce` because it was giving me very dubious results (it would somtimes trigger the function while holding down they key which I believe shouldn't happen with the timeout I gave it). I suppose that's due to the fact that it is more involved than using a timer, or maybe just calculating that based on `reltime` is not a reliable enough method.

I think we should probably move to this approach as a way of debouncing, as it seems more precise.

iddinev:
I'm having some issues with this functionality under vim8, not sure if it has something to do with my configs though..
Tested it with both my vimrc and the min-vimrc provided in the repo (with the addition of the deoplete plugin only).
The strange flashing happens when I scroll through the completion extra info window with the mouse.
Neovim works as in your initial gif.
my vimrc:
Peek 2020-11-12 20-52
min-vimrc:
Peek 2020-11-12 21-38

martskins:
Hm interesting. I'll take a look this weekend. You can disable it meanwhile if it's too distracting

martskins:
Alright, I think this is a limitation of the popup. In vim8 we can't use floating windows, which support scrolling, but I don't think you can scroll on a popup. This could be solved by implementing our own scroll logic (moving the window of lines we display on user scroll). If anyone is willing to work on that that'd be great. Otherwise, I'll just tackle this one when I get some time. Thanks for reporting! @iddinev

iddinev:
But I can scroll the window that gets opened via the hover functionality - is this window different (in type/kind) from the one that this PR opens  when showing this extra completion info ?

Issue #1002, opened 2020-03-29
Support dynamic/overridable selectionUI in textDocument/references


matt-snider:
## Is your feature request related to a problem? Please describe.

The selection UI can only be globally configured but for some operations it would be convenient to be able to use a combination of different UIs.

In particular I have two use cases in mind for `textDocument/references`:

* Jump to a single reference (fzf):

    In this case I just want to see all references and pick one of them, for example because I'm looking for something in particular.

* Explore all references (Quickfix/LocationList):

    In this case I will be looking through many or all of the references, for example because I'm reading through the codebase and trying to understand it.

## Describe alternatives you've considered

I initially examined the code for [textDocument_references][textDocument_references] to see if it accepts
any arguments which might allow this already and concluded that it doesn't.

Then I attempted the following workaround, which doesn't work because the settings seem to be synchronized only at startup:

```viml
" Select a single reference
noremap <leader>sr :call LanguageClient#textDocument_references()<cr>

" Look through all references
noremap <leader>lr :call <SID>list_references()<cr>

function! s:list_references() abort
  " Backup and change selectionUI
  let selectionUI = g:LanguageClient_selectionUI
  let g:LanguageClient_selectionUI = "Quickfix"

  " Call method and restore
  call LanguageClient#textDocument_references()
  let g:LanguageClient_selectionUI = selectionUI
endfunction
```

[textDocument_references]: https://github.com/autozimu/LanguageClient-neovim/blob/8a2396d614579f4b17b33ec942037936b3a73625/src/language_server_protocol.rs#L1723

## Describe the solution you'd like

Ideally `LanguageClient#textDocument_references()` would accept an argument allowing the default (`g:LanguageClient_selectionUI`) to be overidden.

For example:

```viml
" select a single reference - (s)elect (r)eference
noremap <leader>sr :call LanguageClient#textDocument_references()<cr>

" list all references - (l)ist (r)eferences
noremap <leader>lr :call LanguageClient#textDocument_references({'selectionUI': 'Quickfix'})<cr>
```

## Additional context

If you think this seems like a good idea, I could try implementing it, since it seems like it would be a pretty small change.

lrworth:
I want this for exactly the same reason!

runiq:
This could be useful for more than just references.

[...]

PR #1045, opened 2020-06-04
Add function to list computed code lenses

runiq:
I'm trying to get my feet wet by adding a function to expose code lenses from the current state cache.

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1364,6 +1364,15 @@ function! LanguageClient#java_classFileContents(...) abort
     return LanguageClient#Call('java/classFileContents', l:params, l:Callback)
 endfunction
 
+function! LanguageClient#listCodeLenses(...) abort
+    let l:Callback = get(a:000, 1, v:null)
+    let l:params = {
+                \ 'filename': LSP#filename(),
+                \ }
+    call extend(l:params, get(a:000, 0, {}))
+    return LanguageClient#Call('LanguageClient/listCodeLenses', l:params, l:Callback)
+endfunction
+
 function! LanguageClient#handleCodeLensAction(...) abort
     let l:Callback = get(a:000, 1, v:null)
     let l:params = {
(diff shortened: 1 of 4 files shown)


martskins:
It's not clear to me what you are trying to achieve, I thought you wanted to list all the code lenses in the current document, but by your latest commits it doesn't seem like it. Calling `LanguageClient_listCodeLenses` has no associated action, has it? As in, it will just return the code lenses to the client but won't display them in a list or anything like that, is that intended? What would be the use case for this?
If you expand a little more on your intent then I might be able to provide some more thoughts.

Having said that, remember that now PRs should target the `dev` branch, so you may want to change that :smile:

runiq:
> Having said that, remember that now PRs should target the dev branch, so you may want to change that :smile: 

Whoops, sorry, I'll remember that next time!

> What would be the use case for this?

So I'd like to preface this with 'I think this PR is immature' and 'I think I misunderstood LCN's architecture'.

My final goal would be to [integrate LCN with vim-clap](https://github.com/autozimu/LanguageClient-neovim/issues/1002#issuecomment-635473939) so I can select things in clap windows instead of FZF.

I know how to get code actions, reference locations, context menu items, etc. from LCN; the only thing I wasn't able to get were code lenses. I was planning on calling the function in this PR as `:echo LanguageClient_runSync('LanguageClient#listCodeLenses', {})`, seeing how [the denite sources do it that way, too](https://github.com/autozimu/LanguageClient-neovim/blob/4b35f5d6859ab3d42546192cbede9add59a7435a/rplugin/python3/denite/source/codeAction.py#L13-L14).

However, I've realized only today that this is neither necessary nor useful for integrating LCN and clap. I think all I need to do is add a new variant to the `SelectionUI` enum.

All in all, I guess I'm going to close this PR and focus on something like `let g:LanguageClient_selectionUI = function('my_vim_clap_function')`. Sorry for the noise!

martskins:
> add a new variant to the SelectionUI enum.

Exactly, I think this is what you need to do :smile: 

No worries about the "noise", as a user I appreciate more people trying to chip in and contribute to this. :+1:

PR #1047, opened 2020-06-05
Fix recursive path handling for `workspace/didChangeWatchedFiles`

dradtke:
## Problem

`LanguageClient-neovim` currently fails to register filesystem notifications for paths resembling those requested by [eclipse.jdt.ls](https://github.com/eclipse/eclipse.jdt.ls):

```json
"watchers": [
  {"globPattern":"**/*.java"},
  {"globPattern":"**/pom.xml"},
  {"globPattern":"**/*.gradle"},
  {"globPattern":"**/gradle.properties"},
  {"globPattern":"**/.project"},
  {"globPattern":"**/.classpath"},
  {"globPattern":"**/.settings/*.prefs"},
  {"globPattern":"**/src/**"}
]
```

Because none of the requested paths end with `**`, the plugin attempts to register them all as non-recursive file watches, which fails because no file named e.g. `**/*.java` exists.

## Solution

This PR proposes a possible fix, but it's a very blunt one, and I'm open to feedback on how to make it more targeted. Specifically, if a glob pattern contains `**` anywhere, it chops off it and everything after it (defaulting to `.` if it occurs at the beginning of the string, as seen above) and requests a recursive watch; otherwise, it adds a non-recursive file watch.

As is, the 8 watchers requested above all end up adding a recursive `.` watch, which is inefficient. A better solution would be to register recursive watchers on `.` that can still filter events by the rest of the path following `**/`, but I don't know where that would be implemented.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -12,6 +12,7 @@ use crate::{
     viewport,
 };
 use anyhow::{anyhow, Context, Error, Result};
+use glob::glob;
 use itertools::Itertools;
 use jsonrpc_core::Value;
 use log::{debug, error, info, warn};
@@ -2724,13 +2725,27 @@ impl LanguageClient {
                     self.update(|state| {
                         if let Some(ref mut watcher) = state.watchers.get_mut(language_id) {
                             for w in &opt.watchers {
-                                let recursive_mode = if w.glob_pattern.ends_with("**") {
-                                    notify::RecursiveMode::Recursive
-                                } else {
-                                    notify::RecursiveMode::NonRecursive
-                                };
-                                watcher
-                                    .watch(w.glob_pattern.trim_end_matches("**"), recursive_mode)?;
+                                info!("Watching glob pattern: {}", &w.glob_pattern);
+                                for entry in glob(&w.glob_pattern)? {
+                                    match entry {
+                                        Ok(path) => {
+                                            let mode = if path.is_dir() {
+                                                notify::RecursiveMode::Recursive
+                                            } else {
+                                                notify::RecursiveMode::NonRecursive
+                                            };
+                                            debug!(
+                                                "Watching path {} with mode {:?}",
+                                                path.display(),
+                                                mode
+                                            );
+                                            watcher.watch(path, mode)?;
+                                        }
+                                        Err(e) => {
+                                            warn!("Error globbing for {}: {}", w.glob_pattern, e)
+                                        }
+                                    }
+                                }
                             }
                         }
                         Ok(())


martskins:
I think a better approach would be to use `glob`, which is in fact part of the dependencies already. With something like this:

```rust
    for entry in glob::glob_with(fw.glob_pattern.as_str(), MatchOptions::default()).unwrap() {
        // set watcher for entry.unwrap().display()
    }
```

`glob_with` will return a list of paths that match those patterns, and I think it should be able to handle all of the paths in your use case.

This will avoid adding watchers for the same file more than once and also adding watchers for files that the server is not interested (as I understand it, your current implementation would for example watch any extension, not only java, when given the pattern `**/*.java`).

Just a side note, keep in mind that there's #1054 too, which doesn't address this issue in particular but it also involves the watcher logic.

dradtke:
Thanks for the tip, that definitely looks like a better solution. I just pushed an update that uses `glob::glob()` (the default options seem fine), so hopefully that results in better performance.

src/language_server_protocol.rs
```diff
@@ -2675,13 +2676,26 @@ impl LanguageClient {
...
-                                watcher
-                                    .watch(w.glob_pattern.trim_end_matches("**"), recursive_mode)?;
+                                info!("Watching glob pattern: {}", &w.glob_pattern);
+                                for entry in glob(&w.glob_pattern)? {
+                                    match entry {
+                                        Ok(path) => {
+                                            let mode = match path.is_dir() {
+                                                true => notify::RecursiveMode::Recursive,
+                                                false => notify::RecursiveMode::NonRecursive,
+                                            };
```
martskins: Clippy will complain about this, can we change it to

```
let mode = if path.is_dir() {
    notify::RecursiveMode::Recursive
} else {
    notify::RecursiveMode::NonRecursive
};
```

instead?

dradtke: Just pushed an update to change it to use `if/else`.

src/main.rs
```diff
@@ -1,6 +1,8 @@
 #![allow(non_snake_case, non_upper_case_globals)]
 #![deny(clippy::option_unwrap_used)]
 
+extern crate glob;
```
martskins: Ah sorry I missed this one in the previous review. This isn't needed anymore in "rust 2018 edition".

martskins:
Oh and also, PRs should target the `dev` branch now, so please change the target of this PR to that branch.

src/main.rs
dradtke: It was still failing to build without this, do I need to be running a specific version of `rustc`?

martskins: Ah that's because you have imported it in `language_server_protocol.rs` like this `use crate::glob::glob`, which imports the `glob` defined in this project's crate, which is not the external dependency but the one you "re-exported" by adding that `export crate glob`. Dropping this line and changing `use crate::glob::glob` to `use glob::glob` in `language_server_protocol.rs` should do it.

dradtke: Looks like that worked, thanks!

martskins:
@dradtke note that you'll need to rebase this to dev, as it has some merge conflicts now (the code base has drastically changed in the last few days :smile: ). Happy to lend a hand if you need it though.

dradtke:
Just finished rebasing everything on `dev` and the diff looks pretty clean. 🧼

dradtke:
@martskins Any other feedback? It'd be nice to have this PR merged in.

martskins:
Looks good to me, but it'd be nice for @autozimu to take a look, as I'm not sure I've used this feature enough to test it.

dradtke:
@autozimu Any thoughts on getting this PR merged in?

autozimu:
Thanks for the contribution!

PR #1048, opened 2020-06-09
Correctly detect Haskell cabal project root

tomsmeding:
When detecting the project root for a Haskell project, the code looked for a `stack.yaml` or `.cabal` file; however, Cabal project files are called <code>_something_.cabal</code>, not just `.cabal`.

This patch changes the code to properly look for a file with extension `cabal`, not a file named `.cabal`.

Symptom of incorrect behaviour: in a Cabal project, the project root is detected as $HOME instead of the proper project root.


autozimu:
Thanks!

PR #1054, opened 2020-06-15
Wrap notify::Watcher to handle vim's write-via-rename

tomsmeding:
The language server can request a watch on a file that is currently open in vim. If it does so, LanguageClient-neovim adds a file system watcher on that file. Doing this directly via `notify::Watcher` as it did before, however, goes wrong on systems where a file watch is really an inode watch, and vim writes to files by writing to a temporary file and then renaming over the original file. These systems include Linux and macOS.

Related:
- low-levelness of `notify`: https://github.com/notify-rs/notify/issues/247
- my motivation for fixing this: https://github.com/digital-asset/ghcide/issues/618

### DIFF
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -1,0 +1,319 @@
+use anyhow::{anyhow, Result};
+use notify::{DebouncedEvent, RecursiveMode, Watcher};
+use std::collections::{HashMap, HashSet};
+use std::path::{Path, PathBuf};
+use std::sync::mpsc;
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::Duration;
+
+// FSWatch has an outgoing channel on which it pushes events for any of the files or directories it
+// is currently watching. Both directory and file watches are implemented using directory watches,
+// in order to be able to catch write-via-rename tricks like Vim does on save.
+//
+// Thus, FSWatch has a number of directory watches in place. These are all non-recursive watches;
+// recursive watches are connected directly to the FSWatch outgoing channel.
+// There is one thread, with entry point fswatch_service(), that collects all notify events and
+// filters out the ones that we're not interested in. Its incoming channel is attached to the
+// notify::RecommendedWatcher, and the outgoing channel is the outgoing channel of FSWatch.
+
+struct DirWatch {
+    /// whether to also watch the directory itself
+    full_directory: bool,
+    files: HashSet<String>,
+}
+
+fn interested(dirs: &Arc<Mutex<HashMap<PathBuf, DirWatch>>>, path: &Path) -> bool {
+    let dirs = dirs.lock().expect("FSWatch: mutex error");
+
+    if let Some(dw) = dirs.get(path) {
+        if dw.full_directory {
+            return true;
+        }
+    }
+
+    if let Some(parent) = path.parent() {
+        if let Some(dw) = dirs.get(parent) {
+            if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
+                return dw.files.contains(name);
+            }
+        }
+    }
+
+    false
+}
+
+fn fswatch_service(
+    rx: mpsc::Receiver<DebouncedEvent>,
+    tx: mpsc::Sender<DebouncedEvent>,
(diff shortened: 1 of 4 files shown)


src/watcher.rs
```diff
@@ -0,0 +1,316 @@
...
+    rx: mpsc::Receiver<DebouncedEvent>,
+    tx: mpsc::Sender<DebouncedEvent>,
+    dirs: Arc<Mutex<HashMap<PathBuf, DirWatch>>>,
+) -> Fallible<()> {
+    let mut notice_remove = HashSet::new();
+
+    // Policy:
+    // - If we're interested in a path, we handle and push through the event.
+    // - If we're not interested in a path:
+    //   - We _do_ remove the path from notice_remove if applicable; we may have been interested in
```
martskins: Maybe I misunderstood this, but doesn't the remove happen in cases where we're interested as well? At least that's what I get from reading the match arms below.

tomsmeding: You are totally correct, we also remove the path from `notice_remove` in the "interested" case. I'll update the comment to better reflect reality.

src/watcher.rs
```diff
@@ -0,0 +1,319 @@
...
+            None => return Err(failure::err_msg("FSWatch::unwatch on a non-watched path")),
+        };
+
+        let (key, filename): (&Path, Option<&str>) = match info {
+            UnwatchInfo::File(key, filename) => (&key, Some(filename)),
+            UnwatchInfo::Directory(RecursiveMode::NonRecursive) => (path.as_ref(), None),
+            UnwatchInfo::Directory(RecursiveMode::Recursive) => {
+                self.recursive_watcher
+                    .as_mut()
+                    .expect("FSWatch: inconsistent UnwatchInfo")
```
martskins, line 298: I honestly don't understand what the possible consequences of not shutting down the server would be in this case but my first thought would be to handle this case and return an error instead of `expect`ing it. The `expect` calls on the `lock` calls I think are ok, but this one I'm not so sure.

Other than that this looks good to me, although @autozimu might be better suited to review this one.

tomsmeding: For both of the "inconsistent UnwatchInfo" `expect` calls, hitting them would mean that one the one hand we think there should be a watcher set up for the `unwatch`ed path, but on the other hand we find no such watcher in our data structures. This would mean a bug in the (my) watcher code.

If we don't panic there but just return an error, we are apparently in a situation where we're not sure exactly what watchers are or aren't active on what paths. In the end, I believe this would mean that the language server may get file change events that it didn't ask for, or may _not_ get events for paths that it does have watched.

I'm not sure what your policy is in such situations. It might be possible to merge the two parallel data structures (`dirs` and `unwatch_info`), so that this case cannot occur. However, I'm personally not sure exactly how that would play with the `Path` interface; e.g. do `parent` and `file_name` behave predictably and sanely for multiple different ways of writing the same path? That is ultimately why I opted for the `unwatch_info` strategy.

But by all means, if you disagree this can change :)

autozimu: Firstly, thank you for taking time to tackle this.

On the point of `expect` calls, I would prefer to reduce the number of "hard errors" as much as possible. I do agree that in the unexpected case, we can only issuing an error and nothing more. But halting the whole program is a worse user experience, which also hinder future sophisticated error handling such as restarting the watch thread.

I haven't look through the the whole changes yet and I don't know much further on the `parent` and `file_name` topic. Will look further and update here.

tomsmeding: @autozimu Sorry to poll here, but were you intending to do some research and post here again, or are you waiting for me to do anything? I've been happily using my branch for a while now, and it seems to be working (though that doesn't mean the code is perfect, just a confirmation that it's not total crap).

tomsmeding:
Rebased onto latest `dev`, and squashed the two commits into one for good measure.

autozimu:
Merged in via https://github.com/autozimu/LanguageClient-neovim/commit/cccb87b7cb4aadf30d9966e417b0c5ebdfcd2920

Thanks for contributing!

PR #1058, opened 2020-06-18
Fix rust-analyzer runnable actions and create extensions module

martskins:
This PR fixes rust-analyzer runnable actions (runSingle and run), as they were failing due to a change in the arguments provided by rust-analyzer. It also defines types that match the ones used by rust-analyzer in order to avoid abusing `serde_json::Value` and loosing type safety. 

Additionally, this moves the language or language server specific command handling into an `extensions` module for the sake of clarity.


autozimu:
LGTM! Feel free to merge.

PR #1059, opened 2020-06-18
Allow funcref in g:LanguageClient_selectionUI

runiq:
Hot on the heels of https://github.com/autozimu/LanguageClient-neovim/pull/1045#issuecomment-641455349, here's a way to use a selection UI other than FZF. The `FZF` selection UI is implemented as a special backwards-compatible version of the `Funcref` selection UI.

The function has to adhere to the arguments provided to `s:FZF()`, i.e. it has to take `source` and a `sink` arguments. [As in `fzf#run()`](https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfrun), `source` can be either a list or a string, and `sink` may be a string or a reference to another function of one argument (the selected line).

In action, using [vim-clap](https://github.com/liuchengxu/vim-clap) instead of FZF:

works

Config:

```vim
function! MySelectionUI(source, sink) abort
	return clap#run({'id': 'LCN', 'source': a:source, 'sink': a:sink})
endfunction

let g:LanguageClient_selectionUI = function('MySelectionUI')
```

### DIFF
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -118,10 +118,10 @@ function! s:hasSnippetSupport() abort
 endfunction
 
 function! s:getSelectionUI() abort
-	if type(s:GetVar('LanguageClient_selectionUI')) is s:TYPE.funcref
+	if type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
 		return 'funcref'
 	else
-		return s:GetVar('LanguageClient_selectionUI')
+		return get(g:, 'LanguageClient_selectionUI', v:null)
 	endif
 endfunction
 
@@ -210,7 +210,8 @@ function! s:inputlist(...) abort
 endfunction
 
 function! s:selectionUI_funcref(source, sink) abort
-    if g:LanguageClient_selectionUI ==? 'FZF'
+    if get(g:, 'LanguageClient_selectionUI', 'FZF') ==? 'FZF'
+                \ && get(g:, 'loaded_fzf')
         call s:FZF(a:source, a:sink)
     else
         call call(g:LanguageClient_selectionUI, [a:source, function(a:sink)])
@@ -1427,7 +1428,12 @@ endfunction
 
 function! LanguageClient_contextMenu() abort
     let l:options = keys(LanguageClient_contextMenuItems())
-    if type(s:GetVar('LanguageClient_selectionUI')) is s:TYPE.funcref || s:GetVar('LanguageClient_selectionUI') ==? 'FZF'
+    if get(g:, 'LanguageClient_fzfContextMenu', 1)
+            \ && (type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
+                \ || (get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF'
+                    \ && get(g:, 'loaded_fzf')
+                \ )
+            \ )
         return s:selectionUI_funcref(l:options, function('LanguageClient_handleContextMenuItem'))
     endif
 
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -117,6 +117,14 @@ function! s:hasSnippetSupport() abort
     return 0
 endfunction
 
+function! s:getSelectionUI() abort
+	if type(s:GetVar('LanguageClient_selectionUI')) is s:TYPE.funcref
+		return 'funcref'
+	else
+		return s:GetVar('LanguageClient_selectionUI')
+	endif
+endfunction
+
 function! s:useVirtualText() abort
     let l:use = s:GetVar('LanguageClient_useVirtualText')
     if l:use isnot v:null
@@ -199,6 +207,14 @@ function! s:inputlist(...) abort
     let l:selection = inputlist(a:000)
     call inputrestore()
     return l:selection
+endfunction
+
+function! s:selectionUI_funcref(source, sink) abort
+    if g:LanguageClient_selectionUI ==? 'FZF'
+        call s:FZF(a:source, a:sink)
+    else
+        call call(g:LanguageClient_selectionUI, [a:source, function(a:sink)])
+    endif
 endfunction
 
 function! s:FZF(source, sink) abort
@@ -1411,12 +1427,8 @@ endfunction
 
 function! LanguageClient_contextMenu() abort
     let l:options = keys(LanguageClient_contextMenuItems())
-
-    if get(g:, 'loaded_fzf') && get(g:, 'LanguageClient_fzfContextMenu', 1)
-        return fzf#run(fzf#wrap({
-                    \ 'source': l:options,
-                    \ 'sink': function('LanguageClient_handleContextMenuItem'),
-                    \ }))
+    if type(s:GetVar('LanguageClient_selectionUI')) is s:TYPE.funcref || s:GetVar('LanguageClient_selectionUI') ==? 'FZF'
+        return s:selectionUI_funcref(l:options, function('LanguageClient_handleContextMenuItem'))
     endif
 
     let l:selections = map(copy(l:options), { key, val -> printf('%d) %s', key + 1, val ) })
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -1428,7 +1428,9 @@ endfunction
 
 function! LanguageClient_contextMenu() abort
     let l:options = keys(LanguageClient_contextMenuItems())
-    if get(g:, 'LanguageClient_fzfContextMenu', 1)
+    let l:useSelectionUI = get(g:, 'LanguageClient_selectionUIContextMenu',
+				\ get(g:, 'LanguageClient_fzfContextMenu', 1))
+    if l:useSelectionUI
             \ && (type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
                 \ || (get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF'
                     \ && get(g:, 'loaded_fzf')
--- a/autoload/LanguageClient.vim
+++ b/autoload/LanguageClient.vim
@@ -210,11 +210,14 @@ function! s:inputlist(...) abort
 endfunction
 
 function! s:selectionUI_funcref(source, sink) abort
-    if get(g:, 'LanguageClient_selectionUI', 'FZF') ==? 'FZF'
+    if type(get(g:, 'LanguageClient_selectionUI')) is s:TYPE.funcref
+        call call(g:LanguageClient_selectionUI, [a:source, function(a:sink)])
+    elseif get(g:, 'LanguageClient_selectionUI', 'FZF') ==? 'FZF'
                 \ && get(g:, 'loaded_fzf')
         call s:FZF(a:source, a:sink)
     else
-        call call(g:LanguageClient_selectionUI, [a:source, function(a:sink)])
+        call s:Echoerr('Unsupported selection UI, use "FZF" or a funcref')
+        return
     endif
 endfunction
 
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -138,7 +138,7 @@ impl LanguageClient {
             [
                 "!!get(g:, 'LanguageClient_autoStart', 1)",
                 "s:GetVar('LanguageClient_serverCommands', {})",
-                "get(g:, 'LanguageClient_selectionUI', v:null)",
+                "s:getSelectionUI()",
                 "get(g:, 'LanguageClient_trace', v:null)",
                 "map(s:ToList(get(g:, 'LanguageClient_settingsPath', '.vim/settings.json')), 'expand(v:val)')",
                 "!!get(g:, 'LanguageClient_loadSettings', 1)",
@@ -216,7 +216,7 @@ impl LanguageClient {
         let selection_ui = if let Some(s) = selection_ui {
             SelectionUI::from_str(&s)?
         } else if self.vim()?.eval::<_, i64>("get(g:, 'loaded_fzf')")? == 1 {
-            SelectionUI::FZF
+            SelectionUI::Funcref
         } else {
             SelectionUI::default()
         };
@@ -1809,10 +1809,11 @@ impl LanguageClient {
             .collect();
 
         match self.get(|state| state.selection_ui)? {
-            SelectionUI::FZF => {
-                self.vim()?
-                    .rpcclient
-                    .notify("s:FZF", json!([actions?, NOTIFICATION_FZF_SINK_COMMAND]))?;
+            SelectionUI::Funcref => {
+                self.vim()?.rpcclient.notify(
+                    "s:selectionUI_funcref",
+                    json!([actions?, NOTIFICATION_FZF_SINK_COMMAND]),
+                )?;
             }
             SelectionUI::Quickfix | SelectionUI::LocationList => {
                 let mut actions: Vec<String> = actions?
@@ -1843,7 +1844,7 @@ impl LanguageClient {
         let selection_ui_auto_open = self.get(|state| state.selection_ui_auto_open)?;
 
         match selection_ui {
-            SelectionUI::FZF => {
+            SelectionUI::Funcref => {
                 let cwd: String = self.vim()?.eval("getcwd()")?;
                 let source: Result<Vec<_>> = items
                     .iter()
@@ -1852,7 +1853,7 @@ impl LanguageClient {
                 let source = source?;
 
                 self.vim()?.rpcclient.notify(
-                    "s:FZF",
+                    "s:selectionUI_funcref",
                     json!([source, format!("s:{}", NOTIFICATION_FZF_SINK_LOCATION)]),
                 )?;
             }
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -174,11 +174,19 @@ An example implementation is "s:FZF()" in autoload/LanguageClient.vim. For
 details, see the source and sink arguments of fzf#run():
 https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfrun
 
+Another example, this time using vim-clap: >
+
+	function! MySelectionUI(source, sink) abort
+		return clap#run({'id': 'LCN', 'source': a:source, 'sink': a:sink})
+	endfunction
+
+	let g:LanguageClient_selectionUI = function('MySelectionUI')
+
 2.7 g:LanguageClient_selectionUI_autoOpen                *LanguageClient_selectionUI_autoOpen*
 Selection UI auto open when selectionUI is not "fzf"
 
-Valid options: 1 | 0
 Default: 1
+Valid options: 1 | 0
 
 2.8 g:LanguageClient_trace                           *g:LanguageClient_trace*
 
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -312,9 +312,11 @@ Controls how hover output is displayed. Must be one of the following:
 Default: "Auto"
 Valid options: "Never", "Auto", "Always"
 
-2.22 g:LanguageClient_fzfContextMenu         *g:LanguageClient_fzfContextMenu*
+2.22 g:LanguageClient_selectionUIContextMenu *g:LanguageClient_selectionUIContextMenu* *g:LanguageClient_fzfContextMenu*
 
-Should FZF be used for `LanguageClient_contextMenu()`.
+Should the selection UI be used for `LanguageClient_contextMenu()`.
+
+(Alias for the deprecated `LanguageClient_fzfContextMenu`)
 
 Default: 1
 Valid options: 1 | 0
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -163,7 +163,16 @@ Default: 1.
 Selection UI used when there are multiple entries.
 
 Default: If fzf is loaded, use "fzf", otherwise use "location-list".
-Valid options: "fzf" | "quickfix" | "location-list"
+Valid options: "fzf" | "quickfix" | "location-list" | |Funcref|
+
+If you use a |Funcref|, the referenced function should have two arguments
+(source, sink). The "source" argument can be either a command string or a
+list. The sink argument can be a command string or a |Funcref| that takes the
+selected line as its first argument.
+
+An example implementation is "s:FZF()" in autoload/LanguageClient.vim. For
+details, see the source and sink arguments of fzf#run():
+https://github.com/junegunn/fzf/blob/master/README-VIM.md#fzfrun
 
 2.7 g:LanguageClient_selectionUI_autoOpen                *LanguageClient_selectionUI_autoOpen*
 Selection UI auto open when selectionUI is not "fzf"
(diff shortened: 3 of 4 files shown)


src/language_server_protocol.rs
```diff
@@ -1652,9 +1652,10 @@ impl LanguageClient {
             return Ok(result);
         }
 
-        self.vim()?
-            .rpcclient
-            .notify("s:FZF", json!([source, NOTIFICATION__FZFSinkCommand]))?;
+        self.vim()?.rpcclient.notify(
+            "s:selectionUI_funcref",
+            json!([source, NOTIFICATION__FZFSinkCommand]),
```
runiq: `rustfmt` changed indentation here; all I did was to replace `s:FZF` with `s:selectionUI_funcref`.

src/language_server_protocol.rs
```diff
@@ -2060,9 +2061,10 @@ impl LanguageClient {
             .filter(|c| c.is_ok())
             .collect();
 
-        self.vim()?
-            .rpcclient
-            .notify("s:FZF", json!([source?, NOTIFICATION__FZFSinkCommand]))?;
+        self.vim()?.rpcclient.notify(
+            "s:selectionUI_funcref",
+            json!([source?, NOTIFICATION__FZFSinkCommand]),
```
runiq: Same as https://github.com/autozimu/LanguageClient-neovim/pull/1059/files#r442321418.

runiq:
[In action:](https://github.com/liuchengxu/vim-clap/issues/433#issuecomment-646041694)

works

Config:

```vim
function! MySelectionUI(source, sink) abort
	return clap#run({'id': 'LCN', 'source': a:source, 'sink': a:sink})
endfunction

let g:LanguageClient_selectionUI = function('MySelectionUI')
```

src/language_server_protocol.rs
```diff
@@ -171,7 +171,7 @@ impl LanguageClient {
         let selectionUI = if let Some(s) = selectionUI {
             SelectionUI::from_str(&s)?
         } else if self.vim()?.eval::<_, i64>("get(g:, 'loaded_fzf')")? == 1 {
```
martskins, line 173: Should this check for `loaded_fzf` be there? If you are using clap or any other selection UI it doesn't really matter if you have fzf is loaded or not.

autoload/LanguageClient.vim
```diff
@@ -201,6 +209,14 @@ function! s:inputlist(...) abort
     return l:selection
 endfunction
 
+function! s:selectionUI_funcref(source, sink) abort
+    if g:LanguageClient_selectionUI ==? 'FZF'
```
martskins: I think you are missing a check for when the user hasn't set `LanguageClient_selectionUI`. As it stands, if that option is not set, but FZF is loaded, it will use the location list anyways. I would prefer to be explicit and have the user set it to FZF even if it was loaded, but I think we might be better of maintaining the current behaviour so we don't catch anyone by surprise with an unexpected change.

autoload/LanguageClient.vim
```diff
@@ -1401,12 +1417,8 @@ endfunction
 
 function! LanguageClient_contextMenu() abort
     let l:options = keys(LanguageClient_contextMenuItems())
-
-    if get(g:, 'loaded_fzf') && get(g:, 'LanguageClient_fzfContextMenu', 1)
-        return fzf#run(fzf#wrap({
-                    \ 'source': l:options,
-                    \ 'sink': function('LanguageClient_handleContextMenuItem'),
-                    \ }))
+    if type(s:GetVar('LanguageClient_selectionUI')) is s:TYPE.funcref || s:GetVar('LanguageClient_selectionUI') ==? 'FZF'
```
martskins: Same here, if the option is not set at all but FZF is loaded it will use the location list, but the current behaviour is different.

martskins:
Nice work! Left a few comments, but it looks pretty good to me. I would probably add the config you shared here to use clap as an example or documentation, so it's easier for users to find that.

runiq:
I've added the checks for `g:loaded_fzf`.

> I would probably add the config you shared here to use clap as an example or documentation, so it's easier for users to find that.

Right now this depends on @liuchengxu's [patch](https://github.com/liuchengxu/vim-clap/issues/433#issuecomment-646029971) which isn't on master yet; should I just keep this PR open until it's available?

autoload/LanguageClient.vim
```diff
@@ -201,6 +209,15 @@ function! s:inputlist(...) abort
     return l:selection
 endfunction
 
+function! s:selectionUI_funcref(source, sink) abort
+    if get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF'
+                \ || get(g:, 'loaded_fzf')
```
martskins: I don't think this is what you want here. As it stands, if `loaded_fzf` is set then fzf will be used, regardless of what you have configured in `selectionUI`.

I think I would do something like `if get(g:, 'LanguageClient_selectionUI', 'FZF')  ==? 'FZF' && get(g: 'loaded_fzf')` instead, that way FZF will be the default but it will only be used if it's loaded. And if `selectionUI` is another thing, different than FZF, it will go to the else branch.

autoload/LanguageClient.vim
```diff
@@ -1401,12 +1418,10 @@ endfunction
...
     let l:options = keys(LanguageClient_contextMenuItems())
-
-    if get(g:, 'loaded_fzf') && get(g:, 'LanguageClient_fzfContextMenu', 1)
-        return fzf#run(fzf#wrap({
-                    \ 'source': l:options,
-                    \ 'sink': function('LanguageClient_handleContextMenuItem'),
-                    \ }))
+    if type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
+                \ || get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF'
+                \ || get(g:, 'loaded_fzf')
```
martskins: Same here, this will always fall in the if branch if fzf is loaded. I would do something like this.

```vim
if type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
    \ || (get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF' && get(g:, 'loaded_fzf'))
```

martskins:
Added a few more comments on your last changes. As for the patch thing, even though this isn't exactly tied to clap, I would wait for them to merge that patch yeah.

Speaking of tied to clap, I'm not aware of any other selection UI of this sort, but it would be nice to see what else is there, to make sure that we have a generic enough design for this as to allow other plugins to make use of this.

runiq:
> Speaking of tied to clap, I'm not aware of any other selection UI of this sort

There's [skim](https://github.com/lotabout/skim) and its corresponding [Vim plugin](https://github.com/lotabout/skim.vim), at least, but that's pretty much just Clap without the RPC interface. Off the top of my head, I can only think of fzy, but I'm not at a computer at the moment, gotta tuck the young'un in. :)

autoload/LanguageClient.vim
runiq: I just realized I completely dropped the `LanguageClient_fzfContextMenu` check. :flushed: So the final check would be:

```vim
if get(g:, 'LanguageClient_fzfContextMenu', 1)
    \ && (type(get(g:, 'LanguageClient_selectionUI', v:null)) is s:TYPE.funcref
        \ || (get(g:, 'LanguageClient_selectionUI', v:null) ==? 'FZF' && get(g:, 'loaded_fzf')))
```

This is just your version with the `LanguageClient_fzfContextMenu` check in front.
 I'm gonna add that back in, too, and alias it to `LanguageClient_selectionUIContextMenu` or something in a separate commit (naming suggestions welcome, of course).

runiq:
Alright, added in the changes. Sorry for messing that up. :(

runiq:
> Speaking of tied to clap, I'm not aware of any other selection UI of this sort

Here are the ones I've found:

Selector | `source` | `sink`
-|-|-
[fzf](https://github.com/junegunn/fzf) | List of strings \| Vim command as string | Funcref \| Vim command as string
[skim](https://github.com/lotabout/skim) | List of strings \| Vim command as string | Funcref \| Vim command as string
[vim-fzy](https://github.com/bfrg/vim-fzy) | List of strings \| Vim command as string | Funcref
[clap](https://github.com/liuchengxu/vim-clap) | List of strings \| Funcref \| Vim command as string | Funcref \| Vim command as string
[vim-picker](https://vimawesome.com/plugin/vim-picker-may-fear-less) | Shell command as string | Vim command as string
[fzy](https://github.com/jhawthorn/fzy) | None (just a shell command) | None (just a shell command)
[Denite](https://github.com/Shougo/denite.nvim) |  Python | Python
[LeaderF](https://github.com/Yggdroot/LeaderF) | Python | Python
[ctrlp](https://github.com/kien/ctrlp.vim/blob/extensions/autoload/ctrlp/sample.vim) | "The name of the input function including any arguments" as string | "The name of the output function" as string

I've found others, but they all appeared to be variations on the ones above WRT argument usage.

runiq:
Annnd again I realized that I should have based this PR on the dev branch. :( Will rebase.

Edit: Done.

runiq:
The [PR that implements `clap#run`](https://github.com/liuchengxu/vim-clap/pull/46) has been merged now.

autoload/LanguageClient.vim
```diff
@@ -201,6 +209,15 @@ function! s:inputlist(...) abort
     return l:selection
 endfunction
 
+function! s:selectionUI_funcref(source, sink) abort
+    if get(g:, 'LanguageClient_selectionUI', 'FZF') ==? 'FZF'
+                \ && get(g:, 'loaded_fzf')
+        call s:FZF(a:source, a:sink)
+    else
```
martskins, line 218: There's still an issue here, sorry I didn't spot it before. If you set `let g:LanguageClient_selectionUI = 'location-list'` it will error, because it will jump to the else branch and that expects a `Funcref`.

martskins: You'll probably need something similar to what's done in `contextMenu` which opens `inputlist` if it can't use FZF.

martskins:
On an **un**related note, the current `selectionUI` is not really respected everywhere, your PR fixes most of those places, as it calls that new function you implemented instead of directly calling `s:FZF` from Rust. But the contextMenu for example does still not respect that (I know this PR has nothing to do with this :smile: ), since it doesn't really check for the user config but rather it checks that fzf is loaded and uses it if it's available. No need to fix this in this PR, but figured I'd put it in here in case you end up touching that section of the code and come up with a fix for that.

autoload/LanguageClient.vim
runiq: Wow, this is turning out to be much more confusing than I had anticipated. Sorry *again* for messing up.

I guess I'll have to make a table of the possible interactions.

runiq: Wait—that case shouldn't happen at all. `s:selectionUI_funcref()` is only called from the `SelectionUI::Funcref` codepaths in Rust, which would not be hit if `g:LanguageClient_selectionUI = 'location-list'` or `'quickfix'`. Right?

Edit: An error path seems warranted, though.

runiq:
> On an **un**related note, the current `selectionUI` is not really respected everywhere, your PR fixes most of those places, as it calls that new function you implemented instead of directly calling `s:FZF` from Rust. But the contextMenu for example does still not respect that (I know this PR has nothing to do with this smile ), since it doesn't really check for the user config but rather it checks that fzf is loaded and uses it if it's available. No need to fix this in this PR, but figured I'd put it in here in case you end up touching that section of the code and come up with a fix for that.

There's actually `g:LanguageClient_fzfContextMenu` which decides whether FZF or the input list is used for the context menu (FZF is the default, though).

I feel like the selection UI handling is already pretty convoluted at the moment: It needs to deal with backwards compatibility concerns, the Rust—Vim interface, and graceful fallback to sane values in the face of adversity. I'd rather punt this problem to a separate PR if that's okay with you :)

martskins:
> > On an **un**related note, the current `selectionUI` is not really respected everywhere, your PR fixes most of those places, as it calls that new function you implemented instead of directly calling `s:FZF` from Rust. But the contextMenu for example does still not respect that (I know this PR has nothing to do with this smile ), since it doesn't really check for the user config but rather it checks that fzf is loaded and uses it if it's available. No need to fix this in this PR, but figured I'd put it in here in case you end up touching that section of the code and come up with a fix for that.
> 
> There's actually `g:LanguageClient_fzfContextMenu` which decides whether FZF or the input list is used for the context menu (FZF is the default, though).
> 
> I feel like the selection UI handling is already pretty convoluted at the moment: It needs to deal with backwards compatibility concerns, the Rust—Vim interface, and graceful fallback to sane values in the face of adversity. I'd rather punt this problem to a separate PR if that's okay with you :)

Sure thing! :+1:

autoload/LanguageClient.vim
martskins: Ah wait, it's an issue with code actions and code lens actions only. This is rather unfortunate, but maybe I can try fixing that before merging this one, let me see if I can push that through soon enough for you to rebase.

martskins:
@runiq opened #1060 to abstract the selection UI presenting bit. The rebase looks a little painful, but happy to lend a hand if you need it. That fixes the selection UI used by codeLenses and codeActions, so that users that choose to migrate from fzf to clap after this PR don't loose that feature.

I opened it as a draft, as I don't want that to get merged without @autozimu having a look at it. The new function names I'm not 100% happy about, so if you can think of something better, happy to takke suggestions. :smile:

runiq:
Rebased!

So there appear to be two remaining issues: Your [comment above](https://github.com/autozimu/LanguageClient-neovim/pull/1059#discussion_r443008140) about code actions/lenses which should hopefully be addressed now, and the context menu thing:

>> […] the current `selectionUI` is not really respected everywhere […] the contextMenu for example […] doesn't really check for the user config but rather it checks that fzf is loaded and uses it if it's available.
>
> There's actually `g:LanguageClient_fzfContextMenu` which decides whether FZF or the input list is used for the context menu (FZF is the default, though). […] I'd rather punt this problem to a separate PR if that's okay with you :)

After thinking about this, I wouldn't mind including it in this PR after all. :man_shrugging: Using `present_actions()` from the context menu would involve somehow piping through `LanguageClient_contextMenu()` from Vim through to the Rust side, which means adding a new `LanguageClient/presentActions` request if I understand correctly? I'll have to look into this a bit.

I'm a fan of it, though! Always nice to streamline stuff like this. :)

Edit: Actually, it'd probably be a good idea to create those `LanguageClient/present{Actions,List}` requests regardless of where they are needed. Right now, the `present_{list,actions}()` functions are single points of responsibility from the Rust side, but I'm handling option shuffling in multiple points on the Vim side. That feels suboptimal now that #1060 has been merged. I'd rather like for the Vim side to be able to somehow hook into `present_{list,actions}()` as well.

Thoughts?

runiq:
Hmm, I must have messed something up while rebasing. Code actions and lenses don't appear to work—I can select them in Clap, but the action is not triggered. Will investigate.

martskins:
Actually, I completely overlooked something about the context menu, which I found while opening that other PR. There's `g:LanguageClient_fzfContextMenu`, which defaults to `1`, and overrides your `g:LanguageClient_selectionUI` selection, so that explains why `contextMenu` ignores that :smile: . 
Considering that, I'd probably leave it as is, as changing that (even if it was to a "more correct") behaviour would "break" stuff for some users. We can address that eventually though, but it would basically mean removing that `g:LanguageClient_fzfContextMenu` config.

**EDIT**: I just noticed you added a `selectionUIContextMenu` config, that's fine, but we should probably still keep the `fzfContextMenu` around.

runiq:
> I just noticed you added a `selectionUIContextMenu` config, that's fine, but we should probably still keep the `fzfContextMenu` around.

I do keep it around—the code recognizes both, and the help tag for both points to the same doc entry. They just mention that `fzfContextMenu` is deprecated.

martskins:
> Hmm, I must have messed something up while rebasing. Code actions and lenses don't appear to work—I can select them in Clap, but the action is not triggered. Will investigate.

Fwiw, it seems to be working for me.

martskins:
> > I just noticed you added a `selectionUIContextMenu` config, that's fine, but we should probably still keep the `fzfContextMenu` around.
> 
> I do keep it around—the code recognizes both, and the help tag for both points to the same doc entry. They just mention that `fzfContextMenu` is deprecated.

Ah yeah sorry, I missed that.

runiq:
> Fwiw, it seems to be working for me.

Weird. I rebased on top of the base branch, but I'm still not able to trigger *either* code actions or code lenses in Clap. Spent half a day yesterday trying to make things to work, but I had to go deeper and deeper into the rabbit hole—I stopped when I appeared to hit an error in `Runnable::deserialize()`. :( This is the patch that allows me to run code actions at least, but not code lenses:

```diff
diff --git i/src/language_server_protocol.rs w/src/language_server_protocol.rs
index 77b104b..262cbec 100644
--- i/src/language_server_protocol.rs
+++ w/src/language_server_protocol.rs
@@ -1992,9 +1992,9 @@ impl LanguageClient {
             .map(|cl| match &cl.command {
                 None => Err(anyhow!("no command, skipping")),
                 Some(cmd) => Ok(CodeAction {
-                    kind: Some(cmd.title.clone()),
-                    title: cmd.command.clone(),
-                    command: cl.clone().command,
+                    kind: Some(cmd.command.clone()),
+                    title: cmd.title.clone(),
+                    command: cl.command.clone(),
                     diagnostics: None,
                     edit: None,
                     is_preferred: None,
@@ -3418,14 +3418,14 @@ impl LanguageClient {
         let selection: String =
             try_get("selection", params)?.ok_or_else(|| anyhow!("selection not found!"))?;
         let tokens: Vec<&str> = selection.splitn(2, ": ").collect();
-        let title = tokens
+        let kind = tokens
             .get(0)
             .cloned()
-            .ok_or_else(|| anyhow!("Failed to get title! tokens: {:?}", tokens))?;
-        let kind = tokens
+            .ok_or_else(|| anyhow!("Failed to get kind! tokens: {:?}", tokens))?;
+        let title = tokens
             .get(1)
             .cloned()
-            .ok_or_else(|| anyhow!("Failed to get kind! tokens: {:?}", tokens))?;
+            .ok_or_else(|| anyhow!("Failed to get title! tokens: {:?}", tokens))?;
         let actions = self.get(|state| state.stashed_code_action_actions.clone())?;
         let idx = actions
             .iter()
diff --git i/src/types.rs w/src/types.rs
index e82c532..aaa6baa 100644
--- i/src/types.rs
+++ w/src/types.rs
@@ -1252,6 +1252,7 @@ pub struct VirtualText {
 }

 #[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct WorkspaceEditWithCursor {
     pub workspace_edit: WorkspaceEdit,
     pub cursor_position: Option<TextDocumentPositionParams>,
```

Not sure what's going on here, to be honest.

Are you using FZF as your selection UI?

martskins:
Ah, good catch on that missing `rename_all`. I'll push a fix for that in a few minutes.

About the selection UI, I tried FZF and location-list in dev. And I tried both of those and clap in this branch. What are the code lenses you are trying to run? Which language and what do they do?

runiq:
I'm using Rust-analyzer and [this tiny program for testing](https://git.sr.ht/~runiq/seddate/tree/master/src/main.rs). I'm trying to trigger the "Run" codelens from [`fn main()`](https://git.sr.ht/~runiq/seddate/tree/master/src/main.rs#L50).

[These are the interesting bits from a debug log file](https://gist.github.com/runiq/4e7ebdcba6406ded170db0805be564ab#file-lcn-seddate-log-L10-L19). I suspect I'm hitting some edge case from [this recent PR](https://github.com/autozimu/LanguageClient-neovim/pull/1058).

martskins:
Which version of rust-analyzer are you running? `rust-analyzer version` yields `rust-analyzer f4f5117` for me.

runiq:
I had an old version, but the error is the same with the [most recent weekly release](https://github.com/rust-analyzer/rust-analyzer/releases/tag/2020-06-22). The log file was created with that newest release of rust-analyzer.

Edit: The result from `rust-analyzer --version` is `rust-analyzer 31611da`.

martskins:
Ah yeah, I can reproduce now. Seems like there were some other changes as well. This is unfortunate. Let me try and fix that and I'll come back to this. This is not related to this PR though, but it's worth fixing asap anyways.

runiq:
No sweat, shaving yaks is what we do :)

runiq:
I've updated the [log file Gist](https://gist.github.com/runiq/4e7ebdcba6406ded170db0805be564ab)  with a reformatted version of the stashed code actions in line 10; I thought it might be useful for debugging serialization errors.

martskins:
@runiq I've opened #1071. That should allow running both the old and new version of rust-analyzer runnables. It also fixes the issue you found, it seems like I was trying a very narrow set of tests and those were passing but I eventually found the issue you had. I'll merge it in a few minutes anyways, as it's currently broken, so it can only get better :smile:

runiq:
Great, thanks, rebased! Looks like code lenses and actions are triggered just fine now. (I've found some separate small issues, but those are rust-analyzer-related, I believe. I'll file separate issues for those)

<details><summary>Remaining issues (not related to this PR)</summary>
Edit: Spoke too soon. Two remaining issues:

- Applying a code action doesn't reposition the cursor, even though `applyWorkspaceEdit()` appears to finish without errors.
- There seems to be an issue with correct window focus—it's even apparent in the [video in the OP](https://user-images.githubusercontent.com/620981/85034923-e4295400-b182-11ea-9ac1-5a9852ff0a44.gif): After selecting a location, the window doesn't update the line number display on the left, like it's supposed to.

I'll have a look, though I'm not entirely sure yet whether that's an issue with my config, or this PR, or something else…

Edit2: The first issue doesn't seem to have anything to do with this PR, since `applyWorkspaceEdit()` always reports the same position *after* the edit as *before*. Which, [according to the RA manual](https://rust-analyzer.github.io/manual.html#add_impl), shouldn't happen for the actions I tried (impl Struct, add custom impl). I'll file a bug.

Edit3: The second issue is with my vim-numbertoggle installation, and doesn't have anything to do with this either.
</details>

So, at least from my point of view, this PR is finally good to go.

martskins:
I don't fully understand the first issue you mentioned, and I haven't noticed anything weird, but please do file an issue about this so we can track it, if you can add a recording of what's wrong with it it would be great.

As for the PR itself, I'll give this a ride today and see how it behaves, but I'm pretty sure it should be fine now. So if everything works fine I'll merge this at the end(ish) of the day.

Thank you for implementing this. Sorry it took this long and you thank you for bearing with me :smile:

runiq:
> I don't fully understand the first issue you mentioned

If I have this struct (the `|` is the cursor position):

```rust
struct Foo {
bar: u32,|
}
```

and I run the "Implement Foo" code action, I end up with this, which is correct code, but the cursor is in the wrong position (it's supposed to be in the new `impl` block, as per the [RA manual](https://rust-analyzer.github.io/manual.html#add_impl)):

```rust
struct Foo {
bar: u32,|
}

impl Foo {
}
```

martskins:
Ah ok, yeah, that's an experimental feature by rust-analyzer, which is not supported by all clients. See [here](https://github.com/rust-analyzer/rust-analyzer/blob/1bc1f28bc58b2dbcf8f8f548c277e2c90e3075cd/crates/rust-analyzer/src/config.rs#L298).

runiq:
Ah, right, sorry—can't say I fully understand the architecture yet, TBH. :sweat:

PR #1060, opened 2020-06-19
Abstract selection UI and use the correct one for codeActions and codeLenses

martskins:
This PR abstract the creation of quickfix/location list/fzf items into a trait and implements it for the relevant types. This results less repeated code.

This also fixes the selection UI used by codeAction and codeLenses, which was hardcoded to FZF, so nothing else could be used. This PR adds support for location list, and quickfix, and will use the one configured by the user in `LanguageClient_selectionUI`.

There was one place were I couldn't use the new trait to get the list items, and that was in the `text_document_document_symbols` method, I may revisit this later, but if someone wants to give that a try, go ahead.

I'm not exactly excited about the names of the new functions added in this PR, so happy to take suggestions for better ones.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -1420,111 +1420,41 @@ impl LanguageClient {
         let syms = <lsp_types::request::DocumentSymbolRequest as lsp_types::request::Request>::Result::deserialize(&result)?;
         let title = format!("[LC]: symbols for {}", filename);
 
-        let selection_ui = self.get(|state| state.selection_ui)?;
-        let selection_ui_auto_open = self.get(|state| state.selection_ui_auto_open)?;
+        match syms {
+            Some(lsp_types::DocumentSymbolResponse::Flat(flat)) => {
+                self.present_list(&title, &flat)?;
+            }
+            Some(lsp_types::DocumentSymbolResponse::Nested(nested)) => {
+                let mut symbols = Vec::new();
 
-        match selection_ui {
-            SelectionUI::FZF => {
-                let symbols = match syms {
-                    Some(lsp_types::DocumentSymbolResponse::Flat(flat)) => flat
-                        .iter()
-                        .map(|sym| {
-                            let start = sym.location.range.start;
-                            format!(
-                                "{}:{}:\t{}\t\t{:?}",
-                                start.line + 1,
-                                start.character + 1,
-                                sym.name,
-                                sym.kind
-                            )
-                        })
-                        .collect(),
-                    Some(lsp_types::DocumentSymbolResponse::Nested(nested)) => {
-                        let mut symbols = Vec::new();
+                fn walk_document_symbol(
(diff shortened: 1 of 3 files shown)


src/language_server_protocol.rs
```diff
@@ -1708,7 +1634,7 @@ impl LanguageClient {
             },
         )?;
 
-        let response: Option<CodeActionResponse> = serde_json::from_value(result.clone())?;
+        let response = <Option<CodeActionResponse>>::deserialize(&result)?;
```
autozimu, line 1500: Neat!

src/types.rs
```diff
@@ -1094,6 +1099,106 @@ impl ToLSP<Vec<FileEvent>> for notify::DebouncedEvent {
     }
 }
 
+pub trait ListItem {
+    fn quickfix_item(&self, lc: &LanguageClient) -> Result<QuickfixEntry>;
+    fn string_item(&self, lc: &LanguageClient, cwd: &str) -> Result<String>;
```
autozimu, line 1105: I don't have any idea either on the trait naming.

For this particular method though, maybe `stringify` or `to_display`?

autozimu (commented):
Overall, LGTM!

src/types.rs
martskins: Yeah that one is the hardest one. I went with `string_item` to have it named somewhat similarly to `quickfix_item`. To be honest, `to_display` sounds too Rusty to me, I feel like this name should be closer to Vim than to Rust.
The `to_` bit I'm ok with though, but I would probably rename both then, so they are named similarly, `to_quickfix_item` and `to_string_item`/`to_plaintext_item` ?

runiq: What about `to_sel_ui_item`/`to_selui_item`? Or does that look too much like [this](https://cdn.shopify.com/s/files/1/0292/2889/2293/products/Wiggly-Worm.jpg)?

runiq:
> There was one place were I couldn't use the new trait to get the list items, and that was in the `text_document_document_symbol` method, I may revisit this later, but if someone wants to give that a try, go ahead.

[Done](https://github.com/martskins/LanguageClient-neovim/pull/16). (I didn't even know that it was possible to create PRs against PRs)

martskins:
Haven't had any issues with this so far so I'll merge this to dev, if someone finds something please let me know :smile:

src/language_server_protocol.rs
```diff
@@ -2061,38 +1986,44 @@ impl LanguageClient {
...
         }
 
+        let actions: Result<Vec<CodeAction>> = code_lens
+            .iter()
+            .map(|cl| match &cl.command {
+                None => Err(anyhow!("no command, skipping")),
+                Some(cmd) => Ok(CodeAction {
+                    kind: Some(cmd.title.clone()),
+                    title: cmd.command.clone(),
+                    command: cl.clone().command,
```
runiq, line 1996: Err, something happened here and I'm not sure what. :yum:

PR #1063, opened 2020-06-22
Fix doc for diagnostic text settings

110y:
After #1050 has been merged, now we have to use `snake_case` for each `LanguageClient_diagnosticsDisplay` parameters.


110y:
Or should we return to the camel case only about these settings to avoid breaking changes?

martskins:
Ah good catch! I completely missed that one! Just pushed a fix for it to use camelCase for [de]serializing, so we should be alright to close this one now. Thank you again for noting that!

110y:
@martskins 

Thank you for fixing! I've confirmed that!

PR #1065, opened 2020-06-23
Add Plug mappings

martskins:
This PR adds Plug mappings for most functions, this provides slightly cleaner mappings and also easier to remember names and keeps the sanity of not setting default plugins in the user's mapping namespace.

My reasoning behind the names for the mappings was that the user doesn't really have to know or care about all the `textDocument` or `workspace` bit of the commands, so things like `lcn-rename` (instead of `LanguageClient_textDocument_rename`) are friendlier to the user.


autozimu (approved):
Awesome!

PR #1062, opened 2020-06-21
Support Diagnostic.source

110y:
Add `Diagnostic.source` to diagnostic text.

### Background

In case of `gopls`, which is the language server for `Go`, it populates `Diagnostic.source` with the analyzer (linter) name which we violate the rules.
So at least for `gopls`, the `Diagnostic.source` is necessary information for coding.

Fix #296

PR #1073, opened 2020-07-05
Add feature to filter diagnostics messages by its source

110y:
- Add a new option named `g:LanguageClient_diagnosticsIgnoreSources` to filter specific diagnostics message by its `source`.

For example, if we set this new option to following parameters, this language client does not show diagnostics messages which have the source `unsafeptr` or `nonewvars`.
```
let g:LanguageClient_diagnosticsIgnoreSources = ['unsafeptr', 'nonewvars']
```

### Background

Some language servers populate `source` of diagnostics messages. For example, `gopls`, the language server for `Go`, populates `source` with the analyzer (linter) name which we violates.
In case of `gopls`, we can filter some `sources` by options provided by the language server itself. However, there are some `sources` which we could not filter by language server options.
As far as I know, LSP specification does not have capabilities to filter diagnostics information by the source at this time.
Therefore, I think this PR would help those who want to filter diagnostics messages by `source` like me.

Related: #1062


martskins:
I feel like this would be better as a config of each language server, rather than the clients, even if it's not part of the protocol. So I would honestly try and open an issue on gopls or discuss this with the devs in their slack channel first.
But if we were to include this, I would use a map with the language id as a key and the list of sources to ignored as value; to avoid conflicts with sources like `compiler`, for example. Something like this:

```
let g:LanguageClient_diagnosticsIgnoreSources = {
    \ `go`: ['unsafeptr', nonewvars']
    \ }
```

110y:
@martskins 

Thank you for responding!

> this would be better as a config of each language server, rather than the clients, even if it's not part of the protocol.

Yes, I think so too. However, make every language servers implement the feature which is not provided by the LSP spec would be almost impossible.
So, I think it would be worth to introduce filtering feature to the client at this time.

If this can be acceptable, I will change this new option from just an array to a filetype map as you've suggested.
How do you think about?

martskins:
Turns out this is already a feature in `gopls`. If you include something like this in your project's `.vim/settings.json`

```
{
    "initializationOptions": {
       "analyses": {
           "unsafeptr": false,
           "nonewvars": false,
        }
    }
}
```

It will shut down those diagnostic sources. I think this is a better solution, even if it means that all servers must implement it (the alternative is all clients having to implement it), especially because it will directly not run those tools as opposed to running them and hiding the output.

110y:
@martskins 

Yes, I know we can filter some sources by specifying language server options.
However, as I mentioned before, there are some sources which we can not filter by language server options. In case of `gopls`, we can filter `unsafeptr` or `nonnewvars` sources by options, but we can not filter `go mod tidy` source by options. That's why I've created this PR...

autozimu:
LGTM. Thanks!

PR #1075, opened 2020-07-13
Add function and plug mappings to cycle diagnostics

martskins:
This PR creates two new function `LanguageClient_diagnosticsNext` and `LanguageClient_diagnosticsPrevious` which enable the user to jump to the next or previous diagnostic in the current buffer, relative to the cursor's current position.

Peek 2020-07-13 17-56

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -58,6 +58,12 @@ use std::{
     time::{Duration, Instant},
 };
 
+#[derive(PartialEq)]
+pub enum Direction {
+    Next,
+    Previous,
+}
+
 impl LanguageClient {
     pub fn get_client(&self, language_id: &LanguageId) -> Result<Arc<RpcClient>> {
         self.get(|state| state.clients.get(language_id).cloned())?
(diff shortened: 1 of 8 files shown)


martskins:
@autozimu I'm not 100% sure if I missed something and there's already something in place for this or if this is in fact a missing feature, so I'll this for you to comment on and then I'll merge it if there aren't any functions that provide this functionality.

autozimu:
I personally use quickfix to display diagnostics (the default) with `:cp`, `:cn` to jump between locations. While using location list, I believe `:prev`, `:lnext` serve similar purpose. Is this actually sufficient?

martskins:
Not quite. The only bit that I find difficult with that workflow is that the items in the location/quickfix list are (logically) not re-sorted every time you move the cursor, so the "next" item in the location list is not really the next diagnostic based on the position of the cursor, which makes navigation a little difficult if you have many diagnostics. It can also be a little surprising, if you have the location list closed and do `:lnext` then you don't really know which diagnostic that is gonna take you too.

Another thing is that diagnostics do not seem to be ordered by line number in the location/quickfix list, maybe that can be solved though, but it still doesn't solve the previous issue.

autozimu:
Ahh, makes sense. Can you add this gist to the function doc? Then let's
merge it in.

martskins:
I adjusted the doc to make it a little clearer, but not sure if that's what you meant.

src/language_server_protocol.rs
```diff
@@ -603,6 +609,37 @@ impl LanguageClient {
...
+
+            if let Some((_, diagnostic)) = diagnostics.iter_mut().find_position(|it| {
+                if direction == Direction::Next {
+                    it.range.start.line > line
+                } else {
+                    it.range.start.line < line
+                }
+            }) {
+                let target = diagnostic.range.start.line + 1;
+                let _: String = self.vim()?.rpcclient.call("execute", json!([target]))?;
```
autozimu: Is it convenient to jump to proper column as well? Especially helpful when there are more than one diagnostics reported for the same line.

martskins: Yeah good call!

PR #1089, opened 2020-08-26
Force hover contents to display in plaintext

dradtke:
This PR is mostly a proof-of-concept and not intended to be merged as-is, but demonstrates the problem and what a potential solution might be, which is that hover popups can look odd in (Neo)Vim when using markdown.

**[Image comparison](https://imgur.com/a/xLsA9eE)**; the top image is what it looks like today, and the bottom is with this solution applied.

While the protocol and client let you select your preferred markup format between `plaintext` and `markdown`, it appears that the [java language server](https://github.com/eclipse/eclipse.jdt.ls) always returns markdown. While this is a bug in the language server, I realized that (Neo)Vim does not have the capability to properly render markdown anyway; it can syntax-highlight it, but that's not the same as rendering it. Given the limitations of the editor, it may make sense to always simplify hover results by removing links or other markdown noise.


martskins:
Using something like `plasticboy/vim-markdown` and adding this to your config `set conceallevel=2` should hide most of that. So, considering that there's already a way to do this without having to play outside the rules of the protocol and also considering that this is really a server issue, I'll close this PR. But feel free to re-open if you wish to continue the discussion on this.

PR #1090, opened 2020-08-28
support the workspace/configuration request

### LINKED ISSUE #923: TexLab configuration

In order to [configure the texlab language server](https://texlab.netlify.com/docs/reference/configuration):

* texlab only use the `workspace/configuration` request to fetch the configuration from the client.
* LanguageClient-Neovim only use the `workspace/didChangeConfiguration` request to push the configuration to the server.

It would be nice if LanguageClient-neovim could support the  `workspace/configuration` request so that all features of texlab can be used together with LanguageClient-Neovim!

## Additional context

Here is the mirrored feature request made on the TexLab repo: https://github.com/latex-lsp/texlab/issues/123

euclio:
This is needed to send configuration to rust-analyzer and the TexLab language server.

Fixes #923.

### DIFF
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -26,8 +26,8 @@ use lsp_types::{
     CodeActionContext, CodeActionKind, CodeActionKindLiteralSupport, CodeActionLiteralSupport,
     CodeActionOrCommand, CodeActionParams, CodeActionResponse, CodeLens, Command,
     CompletionCapability, CompletionItem, CompletionItemCapability, CompletionResponse,
-    CompletionTextEdit, Diagnostic, DiagnosticSeverity, DidChangeConfigurationParams,
-    DidChangeTextDocumentParams, DidChangeWatchedFilesParams,
+    CompletionTextEdit, ConfigurationParams, Diagnostic, DiagnosticSeverity,
+    DidChangeConfigurationParams, DidChangeTextDocumentParams, DidChangeWatchedFilesParams,
     DidChangeWatchedFilesRegistrationOptions, DidCloseTextDocumentParams,
     DidOpenTextDocumentParams, DidSaveTextDocumentParams, DocumentChangeOperation, DocumentChanges,
     DocumentFormattingParams, DocumentHighlight, DocumentHighlightKind,
@@ -963,6 +963,7 @@ impl LanguageClient {
                     }),
                     workspace: Some(WorkspaceClientCapabilities {
                         apply_edit: Some(true),
+                        configuration: Some(true),
                         did_change_watched_files: Some(GenericCapability {
                             dynamic_registration: Some(true),
                         }),
@@ -1789,6 +1790,23 @@ impl LanguageClient {
             DidChangeConfigurationParams { settings },
         )?;
         Ok(())
+    }
+
+    pub fn workspace_configuration(&self, params: &Value) -> Result<Value> {
+        let config_params = ConfigurationParams::deserialize(params)?;
+
+        let settings = self.get_state(|state| state.initialization_options.clone())?;
+
+        let configuration_items = config_params
+            .items
+            .into_iter()
+            .filter_map(|item| {
+                let section = format!("/{}", item.section?.replace(".", "/"));
+                settings.pointer(&section).cloned()
+            })
+            .collect::<Value>();
+
+        Ok(configuration_items)
     }
 
     pub fn handle_code_lens_action(&self, params: &Value) -> Result<Value> {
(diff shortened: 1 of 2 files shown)


martskins:
I think this makes sense. Could you fix the linting issues that were raised in the CI job? I'll try your PR out today and tomorrow and merge after the lint errors are fixed. :+1:

euclio:
I realized I was merging this PR into the wrong branch, which is why the lint errors were showing up. Fixed.

martskins:
This is breaking `gopls` (the only one I tried so far) for some reason. I still haven't dived in to investigate what's going on, but figured I'd share my finding to let you know why this hasn't been merged yet.

euclio:
I'm not familiar with gopls, but it sounds like there might be some discrepancy between how it and rust-analyzer handle configuration: https://github.com/golang/go/issues/38819

euclio:
Yes, this is an instance of microsoft/language-server-protocol#972.

martskins:
Too bad that discrepancy exists there. I suppose we should wait until that is resolved then, as it could mean breaking the client with some servers.

ccope:
Could it be merged but only enabled with a config flag? It seems useful for some users now, and could be enabled only for file types with compatible servers.

src/language_server_protocol.rs
```diff
@@ -2018,6 +2019,32 @@ impl LanguageClient {
...
+
+        let settings = self.get_workspace_settings(&root)?;
+
+        let configuration_items = config_params
+            .items
+            .into_iter()
+            .map(|item| {
+                let mut settings_val = &settings;
+
+                for item in item.section.unwrap().split('.') {
```
martskins: The `unwrap` here is a little dangerous. I have a few suggestions around this. 

- I'd call `.filter(|item| item.section.is_some())` before the call to `.map()` to have that `unwrap` be safe.
- The unwinding of the section you are doing there can probably be achieve more easily by using `settings.pointer`, I would change the body of that map to something like this:
    ```
    let section = format!("/{}", item.section.unwrap().replace(".", "/"));
    settings
        .pointer(section.as_str())
        .cloned()
        .unwrap_or(Value::Null)
    ```
   Note that, as per the specification, null must be returned if the section is not found in the config.
- this will also no longer return a Result on the map, so no need to collect into `Result<Value>` anymore.

martskins:
I guess we could add a config flag for this and have it disabled by default. If you are happy to add that I think we can make this work @euclio .

Also, left a comment with a change I would make.

martskins:
I was browsing the code for rust-analyzer and it seems like it actually behaves the same as gopls. It expects the initialization options to be under a keyed object with the name of the language server. See https://github.com/rust-analyzer/rust-analyzer/blob/bf84e4958ee31c59e5b78f60059d69a73ef659bb/editors/code/src/client.ts#L40

So I think we're just not picking the correct settings there. We should modify our initialize request to fix this, as it makes `workspace/configuration` make sense when it sends a section with the name of the server. 

I can open a PR to fix that, it will introduce a breaking change, but I suppose it's for a greater good.


**EDIT**: This is a lot trickier than expected, because by the time we issue the initialiize request we don't know the name of the server, so we can't fetch the settings for it. Maybe we want to fetch the section `initializationOptions` when the `workspace/configuration` request requests a section with the same name as the server? Seems a little nasty, but it's an option I suppose, and we don't break compatibility.

src/language_server_protocol.rs
```diff
@@ -2045,6 +2046,27 @@ impl LanguageClient {
...
+        let configuration_items = config_params
+            .items
+            .into_iter()
+            .filter_map(|item| {
+                let section = format!("/{}", item.section?.replace(".", "/"));
+                settings.pointer(&section).cloned()
+            })
+            .collect::<Value>();
+
+        Ok(serde_json::to_value(&configuration_items)?)
```
martskins: `configuration_items` is already a `Value`, so there's no need to `serde_json::to_value` here.

martskins:
Added a new comment. I'm not sure how this would work for you though, as I suppose it would suffer from the same issue that `gopls` and `rust-analyzer` suffer with our plugin. Which is that the initializationOptions are not really in the same form as the settings to be read when the server requests configuration via `workspace/configuration`. It might takes us a little longer to merge this now that we're aware of that issue to be honest.

**EDIT**: Just saw you addressed my comment already.

martskins:
@euclio See #1116, where I propose a way to properly handle initializationOptions, which would enable us to merge this without breaking anything.

martskins:
@euclio I'm planning to merge that PR I linked above (#1116). If you are happy to rebase this one to that one, and adjust this so that it uses `State::initialization_options` instead of `self.workspace_settings`, I can merge this as soon as the other one gets merged. I have tested this already and it seems to work just fine.

euclio:
Rebased.

martskins:
@euclio Sorry to bother again, just merged that branch to dev, but it seems like this needs rebasing again now. Happy to do it myself, but just sending the notice in case you prefer to do it yourself 😁 .

euclio:
Happy to do it, but I won't be able to get to it until tonight.

martskins:
Yeah no worries! Any time is fine 👍

euclio:
Rebased!

martskins:
Thank you!

PR #1093, opened 2020-09-07
Update documentation with Plug mappings

aeolyus:
Update the documentation with Plug mappings from #1065 to match the README

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -36,14 +36,20 @@ like to map these functions to shortcuts, for example: >
     nnoremap <silent> gd :call LanguageClient#textDocument_definition()<CR>
     nnoremap <silent> <F2> :call LanguageClient#textDocument_rename()<CR>
 
+You can also use slightly cleaner Plug mappings instead of the full function name: >
+
+    nmap <silent> K <Plug>(lcn-hover)
+    nmap <silent> gd <Plug>(lcn-definition)
+    nmap <silent> <F2> <Plug>(lcn-rename)
+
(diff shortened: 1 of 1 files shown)


doc/LanguageClient.txt
```diff
@@ -32,18 +32,18 @@ At this point, call any provided function as you like. See
 |LanguageClientFunctions| for a complete list of functions. Usually one would
 like to map these functions to shortcuts, for example: >
```
martskins: This mention about function doesn't make sense anymore if you change the functions for the plug mappings. I think I would keep both, the function example and the plug mappings example, as a way of illustrating the two possibilities. What do you think?

aeolyus: Oh yeah that's a good point!

Maybe we could do something like this?
e.g.
``` vim
At this point, call any provided function as you like. See
|LanguageClientFunctions| for a complete list of functions. Usually one would
like to map these functions to shortcuts, for example: >

    nnoremap <silent> K :call LanguageClient#textDocument_hover()<CR>
    nnoremap <silent> gd :call LanguageClient#textDocument_definition()<CR>
    nnoremap <silent> <F2> :call LanguageClient#textDocument_rename()<CR>

You can also use slightly cleaner Plug mappings instead of the full function name: >

    nmap <silent> K <Plug>(lcn-hover)
    nmap <silent> gd <Plug>(lcn-definition)
    nmap <silent> <F2> <Plug>(lcn-rename)
```
Please suggest better wordings of the "or use Plug mappings" phrase if you have any! Thanks!

martskins: That looks good to me :+1:

PR #1097, opened 2020-09-22
Update LanguageClient.txt

craigmac:
Checking for nvim builtin function 'nvim_open_win' always fails with 0 without asterisk in front of it to tell exists() that it is a function.


autozimu:
Thanks!

Issue #852, opened 2019-07-12
ALE-like underscore/wavy for diagnostics 


Huxpro:
## Is your feature request related to a problem? Please describe.

Currently, if [ALE](https://github.com/w0rp/ale) is not running along with this extension. The diagnostics from LSP only display sign in column and virtual texts but ignore the `Range`.

![Screen Shot 2019-07-12 at 1 12 06 PM](https://user-images.githubusercontent.com/5563315/61155834-bd5c9b00-a4a6-11e9-80b1-84ab42331b46.png)

## Describe the solution you'd like

Have native supports on that. 
Btw, it seems like ALE itself is intended to evolve into a full-fledged language clients as well.
Do we want to depend on it?

## Alternative

As `CoC.vim`? [Show diagnostics via ALE](https://github.com/autozimu/LanguageClient-neovim/issues/800).

## Additional context

I am confusing by the fact that [Updates - Diagnostics](https://github.com/autozimu/LanguageClient-neovim/issues/35#issuecomment-288732042) does show the underline. Does it come from ALE or this extension?

autozimu:
There is no intention for this plugin to depend on other plugins regarding LSP as it is positioned to be the lowest level to talk to language servers.

The underline of GIF showing in Updates is indeed displayed by this plugin.

The styles for highlighting ranges of text are defined as https://github.com/autozimu/LanguageClient-neovim/blob/5491a64a07b259568e26981e5ec0d78ba8cec756/doc/LanguageClient.txt#L95-L125. More specifically, it is `texthl`. You might want to make sure your theme display those highlight groups prominently or customize this setting to different highlight groups.

PR #1101, opened 2020-10-09
Add support for diagnostics display funcref for more flexible integration

### LINKED ISSUE #800: Show diagnostics via ALE
This is based on my experience using coc.nvim. I prefer this LSP client (it's purer than coc.nvim and is FZF-friendly), but coc.nvim has some undeniably cool features that I would like to see being ported here. I might help as well if this is out of scope or too difficult to be implemented, even though my Rust knowledge is limited.

One great feature of coc.nvim is being able to display diagnostics via ALE ([coc's implementation here](https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/manager.ts#L499)). This way, warnings and errors are visually unified and they don't need to override one another.

I'll try to be as technical as possible:
ALE has a method called `ale#other_source#ShowResults` that enables showing results (diagnostics) from another source of information. It receives three arguments: buffer number, source name and a so-called `loclist`.

According to `:h ale-loclist-format`, `loclist` is a dictionary. Here are the relevant fields for this plugin:
* `text` - Required error message.
* `detail` -  Optional, more descriptive message. If set, it is displayed in place of `text` when `ALEDetail` is called.
* `lnum` - Required, preferably a number, as when a string it gets converted using `str2nr()`. It refers to the error line number. Check docs for more details.
* `col` - Optional error column number, default is 0.
* `end_col` - Optional end column number, improves error highlights.
* `end_lnum` - Optional end line number, it improves multi-line error highlights.
* `filename`-  Optional filename. Should be an absolute path to the file. Check docs for more details.
* `type` - Diagnostic severity. Defaults to `'E'`.

Using two diagnostics systems, which is inferior to unifying both. For example, I'd rather have only one Lightline extension than two or more.

## Additional context
I haven't peeked at the source code yet, so I'm not aware of how difficult it is to implement this feature, nor if it's even relevant for the maintainers. Anyway, I thought it would be a great addition to this plugin as ALE is widely adopted for linting and stuff and it's diagnostics system is pretty robust.

Excellent work so far! Thank you very much.

martskins:
This PR provides a config option for a function name to be used as a callback for the publish diagnostics call. What this does is enable us to integrate with external plugins that are used to display diagnostics, such as `dense-analysis/ale`. The documentation includes an example of how to set that up.

Note that the virtual texts will still be displayed if configured.

Fixes #800 and #852.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -211,6 +211,7 @@ pub struct State {
     pub hide_virtual_texts_on_insert: bool,
     pub echo_project_root: bool,
     pub enable_extensions: Option<HashMap<String, bool>>,
+    pub diagnostics_display_funcref: Option<String>,
 
     pub server_stderr: Option<String>,
     pub logger: Logger,
@@ -296,6 +297,7 @@ impl State {
             preferred_markup_kind: None,
             enable_extensions: None,
             code_lens_hl_group: "Comment".into(),
+            diagnostics_display_funcref: None,
 
             logger,
         })
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,4 +1,3 @@
-use crate::logger::Logger;
 use crate::rpcclient::RpcClient;
 use crate::sign::Sign;
 use crate::{
@@ -7,6 +6,7 @@ use crate::{
     vim::Vim,
     watcher::FSWatch,
 };
+use crate::{logger::Logger, vim::Funcref};
 use anyhow::{anyhow, Result};
 use jsonrpc_core::Params;
 use log::*;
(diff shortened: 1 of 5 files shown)


martskins:
This one is a little more involved or disruptive so I'll leave it to you to decide @autozimu, but I think it's a good way to integrate with ale. I think in our case, given that the server publishes the diagnostics when it pleases, we don't need to answer the `ALEWantResults` autocmd, so this should be enough.

src/types.rs
```diff
@@ -211,6 +211,7 @@ pub struct State {
     pub hide_virtual_texts_on_insert: bool,
     pub echo_project_root: bool,
     pub enable_extensions: Option<HashMap<String, bool>>,
+    pub diagnostics_display_funcref: Option<String>,
```
martskins: Maybe we should reuse the `diagnostics_list` config instead? And allow the values to be either `Quickfix`, `Location`, `Disabled` or a `Funcref` ? I think that makes more sense.

martskins: Pushed another commit to remove that config and replace it for support for `funcref` on `LanguageClient_diagnosticsList`. Made some changes to how selectionUI works in those terms as well, for the sake of uniformity.

PR #1106, opened 2020-10-18
Add Config struct and clean up sync_settings function

martskins:
I've been playing around with the idea of setting up a derive for getting the config values more easily. So in this PR I created a library for that derive (`VimEval`). The derive also accepts a field level attribute (`vim_eval`) with three possible meta tags:

- `default`: specifies the default value to be used if the value evaluated from vim ends up being `v:null`.

- `rename`: overrides the name of the variable to fetch with the given value. If omitted it will use the camelCased name of the fields with the prefix `LanguageClient_`, so for example if the field is named `auto_start` the name of the variable to populate that field will be `LanguageClient_autoStart`.

- `eval_with`: this meta tag takes precedence over the other two, meaning that if this one is present, the other two aren't considered. The value for this tag indicates a function name to call (vim-side) to evaluate this value. This is used for example for the `is_nvim` field with the attribute set to `#[vim_eval(eval_with="has('nvim')")]`.

All of that, combined with serde_json, lets us make a single eval call in vim and get a JSON we can deserialize to a Config struct.

The idea behind all this is to make `sync_settings` a little simpler, as it was getting a bit out of control and adding config variables was a bit painful. I know that this is effectively more code than what we had before, but hopefully this makes managing and adding fields to that config easier.

This is just a draft PR, as I mentioned, this is only an experiment, but if you think it's worth it I can polish this a little and make a proper PR out of it. If you believe this is not the path we want to take, all good as well. 👍 

Thoughts @autozimu ?

**EDIT**:

You'll notice I also moved all the config variables to it's own struct, this will also allow us to move the config to be a field of `LanguageClient` instead of `State`. And that in turn enables us to use a `RwLock` for reading the config instead of relying on a `Mutex`, which makes sense, since most of our interactions with the config are reads. I have some progress done on this as well, but I'll probably leave that for a subsequent PR, as it would be a little too much to review if I included that here 😁.


martskins:
Closing in favor of #1156

PR #1108, opened 2020-10-22
Add support for rust-analyzer chaining hints

martskins:
This PR adds support for rust-analyzer chaining hints. rust-analyzer exposes this as a client->server request, so in order for this feature to be able to be turned on/off and have it behave similarly to other clients I had to retain the initialization options within the state of the client and check for the flag before issuing the request.

By default this feature is disabled, as it is a little invasive, but you can enable it by having a workspace settings (`.vim/settings.json`) that looks something like this:

```json
{
    "initializationOptions": {
        "inlayHints": {
            "enable": true
        }
    }
}
```

Note that you can also disable each type of hint individually, which the server then filters, but this `enable` flag is only valid/controlled client-side. The value of this is that it saves us a request to the server and it allows us to disable this as a default.

Also, the hints will only be visible if you are using virtual texts for code lens (`LanguageClient_useVirtualText = 'CodeLens'` or `LanguageClient_useVirtualText = 'All'`). And in the case that you are using virtual texts for both diagnostics and code lens, the diagnostics will take precedence over the hints/code lens.

This is how it looks like:

image

### DIFF
--- a/src/extensions/mod.rs
+++ b/src/extensions/mod.rs
@@ -2,3 +2,43 @@ pub mod clangd;
 pub mod gopls;
 pub mod java;
 pub mod rust_analyzer;
+
+use crate::language_client::LanguageClient;
+use anyhow::Result;
+
+impl LanguageClient {
+    pub fn text_document_inlay_hints(&self, language_id: &str, filename: &str) -> Result<()> {
+        if !self.extensions_enabled(language_id)? {
+            return Ok(());
+        }
+
+        let server_name = self.get_state(|state| match state.capabilities.get(language_id) {
+            Some(c) => c
+                .server_info
+                .as_ref()
+                .map(|info| info.name.clone())
+                .unwrap_or_default(),
+            None => String::new(),
+        })?;
+
+        let hints = match server_name.as_str() {
+            rust_analyzer::SERVER_NAME => self.rust_analyzer_inlay_hints(filename)?,
+            _ => return Ok(()),
+        };
+
+        self.update_state(|state| {
+            state.inlay_hints.insert(filename.to_string(), hints);
+            Ok(())
+        })?;
+
+        Ok(())
+    }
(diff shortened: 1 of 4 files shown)


akhilman:
Can not get it work.

LanguageClient: 7fc2fdd9bb41c15ad330465130f1b8acd19f3168
rust-analyzer: d89c189

Here is log file:
[LanguageClient.log](https://github.com/autozimu/LanguageClient-neovim/files/5426634/LanguageClient.log)

akhilman:
Works fine with rust-analyzer 0d03fe6 (nightly).

martskins:
I also had trouble with d89c189, but it basically wasn't working as expected. It didn't even reply to textDocument/definition. Using e5f252a (which isn't nightly) seems to work though. But yeah, I'd say for this type of things (extensions) there's a high dependency on the server version, as we basically expect a specific response schema, but being an extension, that can change. So the idea is to always support the latest version. If the server introduces a breaking change, we may or may not support the old version, depending on the nature of it.

martskins:
Refactored this a bit in order for it not to fetch the inlay hints on every cursor move. This results in the inlay hints not showing sometimes when you first open a file until you edit that file. That's not ideal but I think it's the lesser of two evils, as calling inlayHints on very big files can result in massive responses from the server and doing that on every cursor move didn't make sense.

Currently the function to calculate the inlay hints is called on `textDocument/didOpen` and `textDocument/didChange`. I decided not to call it in `textDocument/didSave` because usually that's preceded by a `textDocument/didChange`, so it would in most cases result in calling it twice. If anyone can think of other places were it would make sense for us to call it that doesn't result in too many requests do let me know.

--- 

Also found that we weren't filtering the code lenses we send to vim to draw so added the bit we were missing so we reduce the size of the message for `s:set_virtual_texts`.

PR #1111, opened 2020-10-23
Introduce LanguageClient_hoverMarginSize

jez:
The markdown plugin I use has a bunch of regex matching groups that look
for certain characters at the start of a line (like `###`), and the
syntax highlighting breaks when they're not on the start of the line.

We add a single space to the start of every line in a hover comment.
I'd like to make that configurable, so that I can get better markdown
syntax highlighting.

I've chosen to let the user set how big they want the margin to be as an
integer. I'd also be fine with an option that just removes the margin,
but this code is actually shorter to write.

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -635,6 +635,28 @@ Highlight group to be used for code lens.
 
 Default: 'Comment'
 
+2.42 g:LanguageClient_hoverMarginSize        *g:LanguageClient_hoverMarginSize*
+
+When using floating windows to open hover previews (for more information see
+|g:LanguageClient_useFloatingHover|), this setting controls whether and how
+big of a margin to draw around the contents of the hover.
+
+Having some margin looks nice, so the default is 1.
+
+Floating windows don't have a true way to set padding/margin on them, so the
+strategy used is to add one line above and below the language server's raw
+hover response, and one space character before the start of every non-empty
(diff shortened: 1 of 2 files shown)


martskins:
Looks good to me! The PR needs to target the `dev` branch though.

Also, if you could add that configuration variable to the documentation (`doc/LanguageClient.txt`) that'd be great.

Thank you!

doc/LanguageClient.txt
```diff
@@ -635,6 +635,28 @@ Highlight group to be used for code lens.
...
+
+Having a space in front of the hover content can cause some syntax
+highlighters to mess up (because e.g., they have a regex expecting a syntax
+construct to be anchored to the start of a line). To turn off adding this
+margin, set this setting to 0. Example:
+>
+  let g:LanguageClient_hoverMarginSize = 0
+<
+Default: 1
+
```
jez, line 659: Rendered:

Screen Shot 2020-10-23 at 6 58 09 PM

jez:
> The PR needs to target the `dev` branch though

Done, sorry about that.

> Also, if you could add that configuration variable to the documentation (`doc/LanguageClient.txt`) that'd be great.

Done, thanks for the reminder.

@martskins this should be ready to review now

martskins:
Brilliant! Thank you for the contribution!

Issue #774, opened 2019-03-18
Restart language servers when they crash


tbodt:
## Is your feature request related to a problem? Please describe.
I'm running into a bug in clangd that causes it to crash pretty frequently while I edit.

## Describe the solution you'd like
There should be an option to automatically restart a language server that crashes (after printing a message saying it crashed, of course). I think it would also make sense for this to be the default.

## Describe alternatives you've considered
I'm trying to get the clangd bug fixed, but that might take a bit.

afnanenayet:
As an addendum, it might be fruitful to have some sort of metering on this. Say you space out restarts, so you restart after 1 second after the first crash, having the restart time increase after subsequent crashes, or some sort of limit on how many times the server gets restarted.

I'm thinking of a scenario where a langserver just keeps constantly crashing and getting restarted, which could be an issue.

dylanirlbeck:
I second this feature request!

I think a nice alternative would be a command you can call to restart the language server --- say `LanguageClient_restart()`. Then you could just remap a key to this function. Just a thought.

loonies:
Yes please - a command for restarting the server.

It happens to me almost every time due to bug/crash during a debugging session - so annoying.

ulidtko:
+1 on this from me.

I've tried emulating it via `map <Leader>r :LanguageClientStop<CR>:LanguageClientStart<CR>` — but that behaves racy, doesn't work. `E900: invalid channel id` appears frequently.

Mild caution for the implementer: please make sure to employ [exponential backoff][] when doing the restarts. The reason is: not all error-exits of the LSP server are crashes due to bug (plus, many of those won't be "cured" by restart — so, try to avoid making a crash loop) — in my case, the LSP server often gets too big appetite for RAM, and gets **OOM-killed** by the system (to prevent system stall).

It's a PITA that instead of simply hitting `\r` to bring it back up, I have to restart whole NVim.

[exponential backoff]: https://en.wikipedia.org/wiki/Exponential_backoff

martskins:
Closing via #1113

Issue #320, opened 2018-03-03
Fails to recover from lsp server dying; only "broken pipe" errors


jwatzman:
It looks like if the LSP that the plugin has connected to dies for any reason, then `LanguageClient-neovim` does not detect this error case, and just spews "broken pipe" errors until the language support is restarted. I would instead expect the plugin to detect this error case and automatically connect to a new LSP.

I was able to reproduce this across a couple of different environments/languages. I first noticed it with an in-house language/LSP at work (which for reasons I won't go into loves to restart itself). Sadly this one is not public right now so I can't link that one, but I was able to reproduce exactly the same behaviour with rls using `kilall rls` to simulate a crash.

I've gone ahead and filled out the below for the nvim/rls case, though this is quite easy to reproduce.

- neovim/vim version: neovim 0.2.0
- Plugin version (git SHA): 63af88b487aea2c1c73092d949f6b90d713bcab0
- Minimal vimrc:
```
call plug#begin()

Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }

call plug#end()

augroup filetype_rust
    autocmd!
    autocmd BufReadPost *.rs setlocal filetype=rust
augroup END

let g:LanguageClient_serverCommands = {
    \ 'rust': ['rustup', 'run', 'stable', 'rls'],
    \ }
let g:LanguageClient_loggingLevel = 'DEBUG'
nnoremap <silent> K :call LanguageClient_textDocument_hover()<CR>
```
- Language server name/link and version: rls-preview 0.124.0-stable (299a6a9 2017-12-25)
- Reproduction steps (from clean state)
    1. Edit any rust file, e.g., `nvim main.rs`
    2. Press `K` (hover), no error.
    3. In another terminal, `killall rls` to simulate an LSP crash.
    4. Press `K` again (or any other LSP action), get only `Broken pipe (os error 32)` which the plugin fails to recover from.
- `/tmp/LanguageServer.log`: empty
- `/tmp/LanguageClient.log`: quite long, here's the relevant bit at the end (can include the whole thing if you want, though again this is quite easy to reproduce):
```
[...]

PR #1113, opened 2020-10-23
Add support for automatic server restart on crash

### LINKED ISSUE #1110: Check whether language server has crashed

Sometimes language servers crash. This means that LSP features stop working until the user intervenes and fixes the reason why the language server crashed.

I'd like to be notified when the server dies.

I'm fine with any of these:

- a custom autocmd, like `User LanguageServerCrashed`
- a custom call back, like how `LanguageClient_registerHandler` works
  - i think this is preferable, because it means i can get a message too
- a global or buffer-local variable that's set

Again, there's nothing I can do to check this now, so any way to check whether the language server process crashed would be great.

martskins:
This PR adds support for automatic server restart retries on crash.

The way this is implemented it allows the user to specify whether or not to try to restart the server on crash via the option `LanguageClient_restartOnCrash` (which defaults to 1). Additionally, the user can specify a max number of retries via `LanguageClient_maxRestartRetries` (which applies for each languageId separately, and defaults to 5).

The restarts follow a very naive exponential backoff, I could have added a new dependency on a library that implements a better backoff logic, but I think this simple approach should be enough.

Also, when a server crashes, the client will emit a new event `LanguageServerCrashed`.

Note that the attempted restart gets the current file name for the server initialisation, so if for example, your Go server crashes, but you are currently editing a JS file, things might go wrong.

Closes #1110, #320 and #774.

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -657,6 +657,20 @@ margin, set this setting to 0. Example:
 <
 Default: 1
 
+2.43 g:LanguageClient_restartOnCrash          *g:LanguageClient_restartOnCrash*
+
+If enabled, the client will attempt to restart the server on the event that it
+unexpectedly crashes.
+
+Default: 1
+
+2.44 g:LanguageClient_maxRestartRetries          *g:LanguageClient_maxRestartRetries*
+
+Max number of times to attempt to recover from a server crash. Each language
+handles its own count independently.
+
+Default: 5
+
 ==============================================================================
 3. Commands                                           *LanguageClientCommands*
 
@@ -1093,6 +1107,11 @@ This event is triggered when diagnostics changed.
 *LanguageClientTextDocumentDidOpenPost*
 
 Triggered after textDocument/didOpen notification is sent to language server.
+
+6.5 LanguageServerCrashed
+*LanguageServerCrashed*
+
+This event is triggered when a language server unexpectedly quits.
 
 ==============================================================================
 7. License                                             *LanguageClientLicense*
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -214,6 +214,15 @@ impl LanguageClient {
             .as_ref(),
         )?;
 
+        #[allow(clippy::type_complexity)]
+        let (restart_on_crash, max_restart_retries): (u8, u8) = self.vim()?.eval(
+            [
+                "get(g:, 'LanguageClient_restartOnCrash', 1)",
+                "get(g:, 'LanguageClient_maxRestartRetries', 5)",
+            ]
+            .as_ref(),
+        )?;
+
         // vimscript use 1 for true, 0 for false.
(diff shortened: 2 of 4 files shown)


jez:
I haven't yet tried this out, but I will later (maybe tomorrow).

The question I will be trying to answer:

Let's say that the server crashes. I wrote an autocmd to handle `LanguageServerCrashed`, and I set a variable `g:it_crashed` that gets shown in a statusline to indicate as such.

Then after 3 retries, the server process comes back (maybe the user tapped a yubikey, maybe the network came back online, maybe a VPN reconnected, etc.) What information is available for me to know that I can overwrite the `g:it_crashed` variable to start showing a good icon in the statusline?

For example, I believe that LanguageServerStarted is emitted when the `bin/languageclient` process starts, but not when the launguage server spawn happens. I am not sure, but that is what I want to check.

jez:
I believe this also implements the request in #993

martskins:
Note that you can just use `b:LanguageClient_isServerRunning` instead of setting that `g:it_crashed` variable. It won't distinguish between the server having crashed and just not running, but maybe that's enough for you.

And yeah, as you said, the event is emitted when the client starts (it's also `LanguageClientStarted`, not `LanguageServerStarted`). If you want to detected when the server has started you can check the variable I mentioned above, that will let you verify if the server is running for that particular filetype too instead of globally.

Good catch on that other issue!

jez:
> Note that you can just use `b:LanguageClient_isServerRunning`

I am pretty sure you cannot currently do this. Maybe I'm reading the code wrong, but `VIM_IS_SERVER_RUNNING` is only ever set to zero in two places:

- when you switch to a new buffer that doesn't have an active language server started yet
- when `lsp_types::notification::Exit::METHOD` is called

it doesn't get set to zero when server spawns correctly at first and crashes later.

doc/LanguageClient.txt
```diff
@@ -635,6 +635,20 @@ Highlight group to be used for code lens.
...
+
+If enabled, the client will attempt to restart the server on the event that it
+unexpectedly crashes.
+
+Default: 1
+
+2.43 g:LanguageClient_maxRestartRetries          *g:LanguageClient_maxRestartRetries*
+
+Max number of times to attempt to recover from a server crash. Each language
+handles it's own count independently.
```
jez:
```suggestion
handles its own count independently.
```

jez:
> only ever set to zero in two places

oh never mind i see that it's now set in a third by this PR. let me try that out locally and see if it's enough

jez:
restart

This is great, thanks!

martskins:
@jez Spotted an issue with this, not critical though. When you manually want to stop the server by calling `LanguageClientStop` the client will automatically try to restart the server because it thinks the server crashed. Fished this in #1120. I tried it and it all seemed to be working normally now, but if you want to give it a try considering that you have airline thingy going on as well, that'd be great 😁

PR #1115, opened 2020-10-24
Remove extra fields from message

martskins:


martskins:
Was trying to verify if javascript-typescript-langserver was still working after this and I couldn't really find the `meta` field in the first place, but I'll keep that there in case an older version still has it.

As for the extra field that Sorbet sends, @jez if you would be so kind to try if this fixes that for you that would be great. And if that's so I can merge this and cut a new release before more users come across that issue.

jez:
This branch fixes the problem, thanks!

martskins:
I'll merge this one then!

Also, @autozimu if you have any objections about this one we can always roll it back and fix the dependency on jsonrpc-core to version 12 again.

PR #1120, opened 2020-10-29
Add `shutdown` implementation

martskins:
This PR adds support for the `shutdown` request and fixes two issues. The first issue is that we were using `exit` instead of `shutdown` to indicate a server shutdown. The second issue is that shutting down (or exiting) the server would now cause the client to try and restart the server (because of the newly added auto-restart logic), so this checks whether the server was running before running the restart logic to fix that.

I've also modified the `LanguageClientStop` command to issue a `shutdown` instead of `exit` now, and call `exit` as a callback to `shutdown`.


jez:
Nice call! I can test this later.

PR #1137, opened 2020-11-08
Fixup development from source instructions

rski:
CONTRIBUTING.md suggests running `make`, which builds the binary in
release mode. It also suggests setting LanguageClient_devel, so vim ends
up trying to use the debug binary. Fixup that inconsistency and add a
few more tidbits to make the instructions more complete.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,10 +1,29 @@
 # Prerequisites
 
-- [rust] toolchain
+- [rust] toolchain or [nix]
 
 [rust]: https://www.rust-lang.org
+[nix]: https://nixos.org/
 
 # Development
+
+Clone this repo into some place, e.g., `~/.vim-plugins`
+
+```sh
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -1,11 +1,30 @@
...
 
+Clone this repo into some place, e.g., `~/.vim-plugins`
+
+```sh
+mkdir -p ~/.vim-plugins
+cd ~/.vim-plugins
+git clone https://github.com/autozimu/LanguageClient-neovim.git
+cd LanguageClient-neovim
+git checkout dev
+# nix users can nix-shell at this point
```
martskins: I think you missed a `use` here? `nix users can use nix-shell...` ?

rski: I meant it in the sense of can (execute) nix-shell, let me make it explicit

martskins:
Thank you!

PR #1155, opened 2020-11-26
Fix markdown syntax on hover for Vim8

rottencandy:
This PR is similar to #1139, it triggers the `InsertLeave` autocmd to let vim-markdown highlight hover text.
But #1139 only works for neovim's floating windows, this PR is adds the same for vim8's popup windows.


martskins:
Ah nice! Thank you for the contribution!

PR #1156, opened 2020-11-27
Add Config struct and move it from State to an RwLock in LanguageClient

martskins:
This PR is basically the same as #1106 but with a much simpler approach, which I think turns out for the better, as it's much more readable.

Closing #1106 in favor of this one


martskins:
~Added a second commit to this that adds `parking_lot` and replaces all our uses of locks with the ones provides by `parking_lot`. The motivation behind that is that those are more lightweight and enable us to run deadlock detection, which can now be activated by setting `let g:LanguageClient_debugLocks = 1`. That will add a flag to the executable, which will in turn spawn a thread which outputs an error and directs the user to the logs if a deadlock is detected.~

I decided against this. The locks are not a problem right now so switching to parking_lot solves nothing.

PR #1173, opened 2020-12-11
Use github actions instead of travis and circle

martskins:
This PR removes travis and circleci in favour of github actions, the motivation behind this is mostly that travis is now in the process of migrating to travis-ci.com and builds have been terribly slow. I think we could potentially solve it by migrating, but I haven't been able to figure out how to do that, at least not without full permissions on the repo.

I also think github actions are a lot more readable and easier to write, so I believe this would work better for us.


martskins:
@autozimu I think I need you to turn off travis to be able to properly replace it by github actions.

autozimu:
Absolutely. I've turned off both Travis and Circle.

BTW, thank you very much for taking the project forward!

PR #1177, opened 2020-12-23
Add support for server specific handlers

### LINKED ISSUE #1176: Feature Request: new autocmd event `LanguageClientInitialized` for registering handlers before anything happens?

In our language server implementation, we have some customize events that I'd like to register handlers for since it would crash if the language client does not handle these events at all. The tricky part is that it sends some of these custom events at the very beginning of a session.

My understanding is that we can only register handler after the language server is started, so I came up with the following:

```vim
augroup LanguageClient_config
  autocmd User LanguageClientStarted call RegisterEventHandlers()
augroup END
```

However, this doesn't really work since by the time languageClient-neovim triggers the LanguageClientStarted event, the server already sent some of the custom events and crashed since they are not handled.

To solve this issue, I tried to add a new event `LagnuageClientInitialized` right after the server is started and this does register the custom event handlers in time:

```diff
diff --git a/src/language_server_protocol.rs b/src/language_server_protocol.rs
index b65868e..d2ac969 100644
--- a/src/language_server_protocol.rs
+++ b/src/language_server_protocol.rs
@@ -3885,6 +3885,10 @@ impl LanguageClient {
         self.sync_settings()?;
         info!("settings synced");

+        self.vim()?
+            .rpcclient
+            .notify("s:ExecuteAutocmd", "LanguageClientInitialized")?;
+
         let command = self
             .get(|state| state.server_commands.get(&language_id).cloned())?
             .ok_or_else(|| {
```

I wonder if this is acceptable to be merged to this repository. Thanks!

martskins:
This PR adds support for server specific handlers that are specified via config. Other than enabling the user to run certain handlers for a specific server. Becase of this being via config, we can register the handlers before initialization, which ensures all events of that type are handled correctly.

An example configuration to use this would look like this:

```
let g:LanguageClient_serverCommands = {}
let g:LanguageClient_serverCommands.go =  {
			\ 'name': 'gopls',
			\ 'command': ['gopls'],
                        \ 'handlers': {
                        \  'window/logMessage': 'LogProgress'
                        \ }
			\}

```

Note that this is only possible with the newly introduced way of specifying a server command.

Closes #1176

### DIFF
--- a/.vim/settings.json
+++ b/.vim/settings.json
@@ -4,5 +4,16 @@
         "all_targets": false,
         "build_on_save": true,
         "wait_to_build": 0
+    },
+    "rust-analyzer": {
+      "diagnostics": {
+        "disabled": ["macro-error"]
+      },
+      "procMacro": {
+        "enable": true
+      },
+      "cargo": {
+        "loadOutDirsFromCheck": true
+      }
     }
 }
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -78,11 +78,11 @@ To use the language server with Vim's formatting operator |gq|, set 'formatexpr'
 Dictionary to specify the servers to be used for each filetype. The keys are
 strings representing the filetypes and the values are either
   1 - a list of strings that form a command
-  2 - a dictionary with keys name, command and initializationOptions, where:
-    - name: is the name of the server, which should match the root node of the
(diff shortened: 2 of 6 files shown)


martskins:
@fanzeyi if you could try this one out and see if it solves your issue it would be great. I tested it locally and it seems to work just fine, but maybe I missed something for your use-case.

PR #1194, opened 2021-02-08
Implement textDocument/semanticTokens/full

### LINKED ISSUE #1193: Preset LanguageClient_semanticHighlightMaps for known servers

clangd and rust-analyzer (and maybe others) have semantic highlighting, would be nice if it worked out of the box.

As far as I understand, vim has a number of preset highlight classes which are used by the built-in syntax highlighting. They very likely already look great on the user's theme. The default `semanticHighlightMaps` could map to these "built-in" highlight classes.

1. Mapping to generic classes like `LanguageClientType`, `LanguageClientMember`, etc.

    This is less work for the user to set up as it's somewhat generic between different language servers, but it's still not out of the box.

## Additional context

I'm not very familiar with vim syntax highlighting, maybe the built-in classes are not really sufficient?

martskins:
This PR implements the `textDocument/semanticTokens/full` LSP method and removes the old "deprecated" (didn't ever really land as an accepted feature of LSP) implementation.

This also sets a default generic mapping of LSP token types to VIM highlight groups that will be merged with the user-provided language-id-specific semantic tokens mappings config which is used to decide which highlight group to assign to each token.

This is still a draft and there a few things left for this to be considered "working":

- [x] Token modifiers
- [ ] ~Map all token types to a vim highlight group~
- [x] Call the function to get the semantic tokens on file save/change/open/etc
- [x] Decide on whether we want to map to vim highlight groups or create our own LCN specific highlight groups that we can link to VIM's generic ones.
- [x] Update user documentation to reflect the removal of textDocument/semanticHighlight and the addition of this one. Also might be a good idea to add a link to the LSP documentation listing all the different token types if we end up supporting all of them.
- [ ] ~Support for `textDocument/semanticTokens/range` (this I believe we could leave out of the first release of this feature)~
- [ ] ~Support for `textDocument/semanticTokens/full/delta` (this could also be optional for the first release)~

Closes #1193

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -97,6 +97,7 @@ pub struct Config {
     ///     { "name": "type", "modifiers": [], "highlightGroup": "Type" }
     /// ]
     pub semantic_token_mappings: Vec<SemanticTokenMapping>,
+    pub semantic_highlighting_enabled: bool,
 }
 
 impl Default for Config {
@@ -134,6 +135,7 @@ impl Default for Config {
             restart_on_crash: true,
             max_restart_retries: 5,
             semantic_token_mappings: vec![],
+            semantic_highlighting_enabled: false,
         }
     }
 }
@@ -172,6 +174,7 @@ struct DeserializableConfig {
     restart_on_crash: u8,
     max_restart_retries: u8,
     semantic_token_mappings: Vec<SemanticTokenMapping>,
+    semantic_highlighting_enabled: u8,
 }
 
 impl Config {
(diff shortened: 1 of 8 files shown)


src/config/mod.rs
```diff
@@ -59,22 +59,31 @@ pub struct Config {
...
+    /// any other mappings included for specific language ids will be merged with the default one
+    /// before highlighting the document.
+    ///
+    /// For example:
+    ///
+    /// {
+    ///     "rust": { "type": "Type", "struct": "Structure" },
+    ///     "go": { "type": "Type", "struct": "Structure" },
+    /// }
+    pub semantic_token_mappings: HashMap<String, HashMap<String, String>>,
```
YaLTeR: I wonder if using the language as a key is the best approach, since:
1. There are multiple languages with the same LS (C and C++ are both clangd).
2. There are multiple LSs for the same language (for C++ there's clangd and ccls).

Maybe using the language server as a type somehow makes more sense? Or if they all use similar semantic token types just put them all in one map without dividing into languages?

martskins:
Yeah I did consider having them keyed by language server name, but considering that we have all other configs based on language id it felt more appropriate. Having said that, it does feel like having a single map for all of them might be a better idea, if users eventually want more flexibility and want to break it down by language server we can do so and support both anyways, but might as well keep it simple for now.

src/config/mod.rs
martskins: Yeah I did consider having them keyed by language server name, but considering that we have all other configs based on language id it seemed more appropriate to use that instead. Having said that, it does feel like having a single map for all of them might be a better idea, if users eventually want more flexibility and want to break it down by language server we can do so and support both anyways, but might as well keep it simple for now.

martskins:
I think the way to go here is use standard vim highlight groups, not LCN defined ones. If users want to define their own groups that's fine, but we have no reason to over engineer this I think. I'm gonna assume we take that path unless someone has a compelling reason not to do so.

martskins:
For the sake of releasing this faster I'm gonna leave the default mappings of LSP tokens to vim highlight groups for a later PR. I'm a little undecided on what the proper mappings are so I guess it could be useful to get some more eyes on that front.

Issue #1007, opened 2020-04-09
textDocument_definition option "gotoCmd" does not work with fzf


marene:
- Did you upgrade to latest plugin version?
commit: `027f574eb58ab33171c3b1b9b2738655f3f60c77`

- Did you upgrade to/compile latest binary? Run shell command
```
?> bin/languageclient --version
languageclient 0.1.156 027f574eb58ab33171c3b1b9b2738655f3f60c77
```
- (Neovim users only) Did you check output of `:checkhealth LanguageClient`?
```
health#LanguageClient#check
========================================================================
  - OK: binary found: /Users/marene/.config/nvim/bundle/LanguageClient-neovim/bin/languageclient
  - OK: languageclient 0.1.156 027f574eb58ab33171c3b1b9b2738655f3f60c77
  - OK: Floating window is supported and will be used for hover
```
- Did you check [troubleshooting]?

[troubleshooting]: https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#troubleshooting

## Describe the bug
When a call to `LanguageClient#textDocument_definition({'gotoCmd': 'split'})` is made and several targets are found, an fzf pane is opened.
After selecting the desired target, the file is opened in the same window instead of being opened in a split

## Environment

- neovim/vim version (`nvim --version` or `vim --version`): `NVIM v0.4.3`
- This plugin version (`git rev-parse --short HEAD`): `027f574`
- This plugin's binary version (`bin/languageclient --version`): `languageclient 0.1.156 027f574eb58ab33171c3b1b9b2738655f3f60c77`
- Minimal vimrc content (A minimal vimrc is the smallest vimrc that could
  reproduce the issue
```
call vundle#begin('/Users/marene/.config/nvim/bundle')

Plugin 'VundleVim/Vundle.vim'
Plugin 'junegunn/fzf', { 'do': './install --bin' }
Plugin 'junegunn/fzf.vim'

Plugin 'autozimu/LanguageClient-neovim'
Plugin 'Shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }

call vundle#end()

set mouse=a
nu
set rtp+=~/.fzf

nnoremap <F12> :call LanguageClient#textDocument_definition({"gotoCmd": "tabedit"})<CR>

" add whatever language server you need
let g:LanguageClient_serverCommands = {
[...]

PR #1195, opened 2021-02-12
Add support for fzf_action in s:FZF

### LINKED ISSUE #629: open buffer in a split or tab when selecting references in fzf window
In default fzf window(eg. created by calling `:FZF `), pressing `ctrl-v` can open file in a new vertical splitted  window. But in fzf window created by language client's references function, this doesn't work. Is there anything different ?

Respect fzf's option `g:fzf_action` and support opening buffer in a split or tab.

cheuk-fung:
Fixes #629, also solves #1007 in a different manner

### DIFF
--- a/doc/LanguageClient.txt
+++ b/doc/LanguageClient.txt
@@ -228,10 +228,11 @@ Valid options: "fzf" | "quickfix" | "location-list" | |Funcref|
 If you use a |Funcref|, the referenced function should have two arguments
 (source, sink). The "source" argument can be either a command string or a
 list. The sink argument can be a command string or a |Funcref| that takes the
-selected line as its first argument.
+selected line as its argument.
 
-An example implementation is "s:FZF()" in autoload/LanguageClient.vim. For
-details, see the source and sink arguments of fzf#run():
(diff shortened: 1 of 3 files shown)


doc/LanguageClient.txt
```diff
@@ -228,10 +228,11 @@ Valid options: "fzf" | "quickfix" | "location-list" | |Funcref|
 If you use a |Funcref|, the referenced function should have two arguments
 (source, sink). The "source" argument can be either a command string or a
 list. The sink argument can be a command string or a |Funcref| that takes the
-selected line as its first argument.
+selected line as its argument.
```
cheuk-fung, line 231: With `sink*` instead of `sink` being passed into `fzf#run(fzf#wrap..`, the selected line is actually the second argument to the sink. The first argument is the `action` users took in the fzf window. The arguments and their order that the `sink` takes should depend on the implementation of the referenced function (`Funcref`).

martskins:
Thank you for your contribution!

Just for the record, I'm merging this as the test has been failing for a while now (trying some stuff to fix that) and the lints I'm already working on a branch that fixes those (although I might open a different PR just for that).

Issue #1198, opened 2021-02-20
Is visualCodeAction working?


weigangd:
## Describe the bug

I am trying to use the "extract function assist" function from rust-analyzer: https://rust-analyzer.github.io/thisweek/2021/02/08/changelog-63.html#new-features . I found the [visualCodeAction](https://github.com/autozimu/LanguageClient-neovim/pull/1022) but I can't make it work. Whenever I select code and use directly `:call LanguageClient#textDocument_visualCodeAction()` nothing happens. Also when I select code and use the context_menu `:call LanguageClient_contextMenu()` and the CodeAction also nothing happens. Codeactions without visual selection work fine.

## Environment

- neovim/vim version : NVIM v0.4.4
- This plugin version : plugs['LanguageClient-neovim'].commit = 'a42594c9c320b1283e9b9058b85a8097d8325fed'
- This plugin's binary version : languageclient 0.1.161

- Language server link and version: rust-analyzer 7435b9e : https://github.com/rust-analyzer/rust-analyzer/releases/tag/2021-02-15

## Additional context

"Extract function assist" works in nvim-lsp, so it shouldn't be a problem of rust-analyzer.

martskins:
It's working for me with that version of rust-analyzer.

![Peek 2021-02-23 19-38](https://user-images.githubusercontent.com/4250565/108898202-d4fe7e00-760e-11eb-8420-7e7589096c0f.gif)


Can you try setting the log level to `DEBUG` and the log after trying this?

weigangd:
Thank you for your answer. Thanks to you GIF i noticed that it works when I use the character visual mode **v** but when I use the line visual mode **V** then nothing happens. I think the important logging lines are:
```
121:37:50 DEBUG reader-None src/rpcclient.rs:207 <= None {"id": 1, "jsonrpc": "2.0", "method": "textDocument/codeAction", "params": {"bufnr": 1, "filename": "/Users/david/projects
/rust/hello/src/main.rs", "character": 0, "handle": true, "languageId": "rust", "line": 3, "range": {"end": {"character": 2147483646, "line": 6}, "start": {"character": 0, "line"
: 3}}}}                                                                                                                                                                           
21:37:50 INFO unnamed src/language_server_protocol.rs:1318 text_document_code_action; params=Object({"bufnr": Number(1), "character": Number(0), "filename": String("/Users/david/
[...]

PR #1199, opened 2021-02-24
Fix get_visual_selection for visual line mode (#1198)

weigangd:
I can't really run the tests because I don't have Docker on my private machine. But in my nvim the changes work fine.


martskins (approved):

martskins:
Thank you for contributing!

PR #1218, opened 2021-04-30
update the tag stack for jump to definition

### LINKED ISSUE #517: Feature request: When jumping to a definition, increment the tag stack

iamcodemaker:
When jumping to a definition, updating the tag stack would be a useful feature. This allows navigating between definitions independently of other code navigation (`ctrl-o`/`ctrl-i` based jumping). Popping the stack is done using `ctrl-t` and works the same as jumping works with a tags file generated by ctags. The logic follows [this example](https://vimhelp.org/tagsrch.txt.html#tagstack-examples).

Closes #517.

#### Considerations
- do we need to check for the `gettagstack` and `settagstack` functions before attempting to use them? They are recent additions to vim.
- this has not been tested with nvim (only vim8)
- if there are multiple potential matches and a selection must be made, the tag stack will not be updated


src/vim.rs
```diff
@@ -275,4 +281,11 @@ impl Vim {
     pub fn set_signs(&self, filename: &str, signs: &[Sign]) -> Result<i8> {
         self.rpcclient.call("s:set_signs", json!([filename, signs]))
     }
+
+    pub fn update_tagstack(&self, winnr: Winnr, item: TagStackItem) -> Result<()> {
```
martskins, line 285: I'm pretty sure `settagstack` is a `neovim` only function, so I think it would be a good idea to make this return early if it's running `vim`.

martskins:
I'm a little confused, I was expecting this to enable both `:tag` and `:pop` to work, and while I seem to be able to jump back in the tag stack using `:pop` I don't seem to be able to use `:tag` to go forwards in the tag stack. Isn't that how it's supposed to work? (genuinely  asking, not 100% sure about how this works)

src/vim.rs
iamcodemaker: This works in recent versions of vim8, including the one I am running.

iamcodemaker:
@martskins I noticed that as well. I'm not sure why `:tag` doesn't work either.

Edit: one would think`:tag` would just walk up the tag stack, but it gives an error message saying the tag is not found. It may have something to do with how I'm setting the tag name. I might have to take a look at the vim source to see what's going on.

iamcodemaker:
Ok, I think I know what's happening. Effectively, this is a limitation (or bug) in the way the tag stack is implemented in vim. When the `:tag` command is run, vim doesn't just jump to the next tag on the stack, it attempts to [find that tag again in the tag file](https://github.com/vim/vim/blob/b7c978154e0816f4dcfae8a06b4ba1bfb7f796f6/src/tag.c#L561). Since there is no tag file the tag not found message is returned.

The tag stack contains the name of a tag and where we jumped _from_ to get to that tag, but it doesn't contain the location of the tag, only the tag name. So it seems the only way to actually jump to the next tag on the tag stack would be to locate that tag again. In the case of `settagstack` the actual jump is done elsewhere and the location of the tag is not know to the tag stack.

src/vim.rs
martskins: ah fair enough! :+1:

iamcodemaker: We may still need to add checks for users of older versions. I'm not sure how to do that yet tho.

martskins:
The tagstack does contain the buf number and the position though, I would have expect it to use that to jump to the next tag. I'll try poking around some other implementations of this to see how they behave, I think `vim-go` has support for tags, so that'll be my first target.

src/vim.rs
martskins: You can use `exists('*settagstack')` for that. Might be worth creating a `SetTagStack` function to do all the getting the current stack, clearing it, and setting the new one plus checking if it's supported all in one function and one call from rust to vim too.

iamcodemaker: That's what this function does already, so I'd only need to add those `exists()` calls. I can figure out how to do that.

iamcodemaker:
> The tagstack does contain the buf number and the position though

No, it doesn't. It only contains that info for where we jumped _from_ (which could be anywhere). It stores the bufnum, matchnum, and tagname of the current tag, which is enough to look it up from the tags file, but there is no tags file if settagstack was used. That's how vim works at least (not sure about nvim).

src/vim.rs
martskins: What I mean is, the function you've set up now issues two rpc messages to vim, one for `gettagstack` and another one for `settagstack` after clearing and setting the new stack. Adding the `exists` call will be a third rpc message sent to vim, so I'm wondering if it might be worth creating a vimscript function that does this instead and calling that instead, so that there's a single rpc message being sent to vim, and there's less network overhead for something that should be relatively simple otherwise. Maybe it's not worth the code complexity though, but worth keeping in mind if it ends up doing more things that require sending extra messages.

martskins:
Ooh ok, I misunderstood that part, yeah, I see now. Thanks for clarifying.

dimbleby:
#917 was a previous try at this - with much the same defects.  There's some relevant discussion in that MR.

Issue #934, opened 2019-12-23
Inline type inference


jestarray:
![image](https://user-images.githubusercontent.com/34615798/71373448-06dc0680-256c-11ea-8eb1-95e8edcfc576.png)

Is there a way to get the type inference to be displayed inlined like vscode? Or is this not supported?

jestarray:
Someone said this could be implemented with "conceal text".

theHamsta:
The inlay hints are requested by a custom request (`rust-analyzer/inlayHints`)
https://github.com/rust-analyzer/rust-analyzer/blob/17dda0972a68dd88a766c223390317dc2cb3ea00/editors/code/src/inlay_hints.ts#L135-L148

I think conceal can only hide text. I think we need to wait until virtual text gets extended in Neovim to be displayed not only in at the end of line but also in the middle. I think this was planned when merging the virtual text commit.

akhilman:
Is it possible to have at least an annotation at the end of the line like CoC has?
![image](https://user-images.githubusercontent.com/180812/96906927-b2bfec80-14c4-11eb-816f-ded4cc3fae58.png)

Also can someone point to the neovim issues about virtual text?

martskins:
@akhilman I think that would be fairly simple to implement. It's very specific to rust-analyzer, but we handle extensions already so I suppose we could add this one as well.

martskins:
@akhilman see #1108 😄 

If you could check out my branch, build from source and verify that it works as you'd expect that'd be great. I'll merge after I giving it a more in-depth try but I think it should be enough.

ibraheemdev:
Was support added in #1106? I can't get it to work with the following config:
```vim
let g:LanguageClient_serverCommands = {
     \ 'rust': {
     \   'name': 'rust-analyzer',
     \   'command': ['rust-analyzer'],
     \   'initializationOptions' : {
     \       'inlayHints': {
     \           'enable': v:true,
     \           'chainingHints': v:true
     \       }
     \   },
     \ },
     \}
```

martskins:
@ibraheemdev it works for me with that exact configuration. Which version of rust-analyzer and language-client-neovim are you using?

ibraheemdev:
@martskins I built rust-analyzer from master and languageclient-neovim from `next`:
```vim
Plug 'autozimu/LanguageClient-neovim', {
   \ 'branch': 'next',
   \ 'do': 'bash install.sh',
   \ }
```
[...]

PR #1221, opened 2021-05-07
Add support for rust-analyzer type & parameter hints

Logarithmus:
Temporary workaround for https://github.com/autozimu/LanguageClient-neovim/issues/934, until `neovim` anticonceal https://github.com/neovim/neovim/pull/9496 is finished.
ChainingHints - `Iterator<u32>`
TypeHints - `: u32`
ParameterHints - `𝑓(a, b, c)`

Demo:

image

### DIFF
--- a/src/config/server_command.rs
+++ b/src/config/server_command.rs
@@ -81,14 +81,14 @@ mod test {
 
     #[test]
     fn test_name_from_command_handles_binary_name() {
-        let name = ServerCommand::name_from_command(&vec!["gopls".into()]);
+        let name = ServerCommand::name_from_command(&["gopls".into()]);
         assert_eq!(name.as_str(), "gopls");
     }
 
     #[test]
     #[cfg(not(target_os = "windows"))]
     fn test_name_from_command_handles_binary_path() {
-        let name = ServerCommand::name_from_command(&vec!["/path/to/gopls".into()]);
+        let name = ServerCommand::name_from_command(&["/path/to/gopls".into()]);
         assert_eq!(name.as_str(), "gopls");
     }
 }
--- a/src/extensions/rust_analyzer.rs
+++ b/src/extensions/rust_analyzer.rs
@@ -1,8 +1,13 @@
-use crate::types;
-use crate::{language_client::LanguageClient, types::WorkspaceEditWithCursor, utils::ToUrl};
+use crate::{
+    language_client::LanguageClient,
+    types::{self, WorkspaceEditWithCursor},
+    utils::ToUrl,
+};
 use anyhow::{anyhow, Result};
+use core::iter;
 use jsonrpc_core::Value;
 use lsp_types::{request::Request, Command, Location, Range, TextDocumentIdentifier};
+use regex::{Captures, Regex};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
@@ -47,14 +52,14 @@ enum GenericRunnableKind {
     Cargo,
 }
 
-#[derive(Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 pub enum InlayKind {
     TypeHint,
-    ParameterHint,
     ChainingHint,
+    ParameterHint,
 }
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct InlayHint {
     pub range: Range,
     pub kind: InlayKind,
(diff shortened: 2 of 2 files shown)


martskins:
Nice! Very clean and readable! :+1:

Logarithmus:
@martskins ready for review I think.

doc/LanguageClient.txt
```diff
@@ -700,6 +674,117 @@ Default: []
...
+        },
+        "rust-analyzer": {
+          "inlayHints": {
+            "enable": true,
+            "chainingHints": true
+          },
+        },
+    }
+
+3.1 rust-analyzer     *LanguageClientSettings_rust-analyzer*
```
martskins: I think there's a misuderstanding here, `settings.json` is not for the settings of `LanguageClient-neovim` extensions, it's essentially the initialization options for the server, so what that file ends up looking wlll not be something we control. So, having said that, I don't think this is worth documenting. We have no control over which settings make sense and which don't, that's up to each specific server.

I honestly thing the whole `Settings` section isn't worth getting into. I think it's fine to give a few examples maybe, but we don't want to start documenting specifics for each server, because we don't have control over them.I would of course keep the explanation of what the settings.json file is that was already there though.

I think the confusion may arise from the fact that we are using those settings as well to enable some calls to the server and whatnot, so we are essentially using them as settings to our extensions as well, but that's purely for simplicity and to keep it transparent for the user.

martskins:
I'm not sure the prefixes for the different hints are easy to remember. I was going back and forth between your comment here on this PR and the screenshot to understand what each hint meant. I'm thinking if maybe surrounding the type parameters with parentheses and leaving the others without a prefix would be clearer ? (and keep the pipe in between the hints)

Nice work though, and thank you for contributing!

doc/LanguageClient.txt
Logarithmus: >I think the confusion may arise from the fact that we are using those settings as well to enable some calls to the server and whatnot, so we are essentially using them as settings to our extensions as well, but that's purely for simplicity and to keep it transparent for the user.

OK, how the user is supposed to learn about the settings of our extensions? Only by looking at the source code?
If you don't want to put this info in a separate `Settings` section, then suggest where it belongs better.

martskins: I think I'm not communicating what I want to communicate correctly. I of course don't expect the user to look at the source code to figure out about the hints. The hints are a feature offered by rust-analyzer in this case, and they are documented there, we don't have any specific toggles for them because they are very specific to rust-analyzer, and we can use the initialization options provided by the user (in `settings.json`) as a config for our side of the feature anyways.

The initialization options, and therefore settings.json file, is a little bit of an obscure corner of the protocol precisely because each server "expects" different values in there. And I wouldn't want to go on the rabbit hole of adding each and every option for each and every server, that's up for each server to document.

Logarithmus:
> I'm not sure the prefixes for the different hints are easy to remember. I was going back and forth between your comment here on this PR and the screenshot to understand what each hint meant. I'm thinking if maybe surrounding the type parameters with parentheses and leaving the others without a prefix would be clearer ? (and keep the pipe in between the hints)

The problem is that Github interpreted those arrows as emojis.

My logic behind those prefixes is quite simple (hopefully, it makes sense):
 - left arrow symbolizes assignment to a variable (from right to left);
 - down arrow symbolizes iterator chain going from top to bottom;
 - right arrow symbolizes input (into function).

I propose the following alternative scheme:
`ChainingHints` - simply `Iterator<u32>`
`TypeHints` - `: u32`
`ParameterHints` - `𝑓(a, b, c)`

What do you think?

doc/LanguageClient.txt
Logarithmus: >And I wouldn't want to go on the rabbit hole of adding each and every option for each and every server, that's up for each server to document.

Of course we aren't going to document each language server's initialization options. The purpose of `Settings` docs section is purely for things like toggle displaying different kinds of inlay type hints, customize prefixes for them, etc. (the things which are managed on our side, not the language server's).

Another solution would be to move these toggles to vim variables of the form `g:LanguageClient_settings_*`.

martskins: >  toggle displaying different kinds of inlay type hints,

This is controlled by the server initialization options. We merely react to the value of the setting by doing stuff client-side so that the setting that we passed on to the server makes sense.

>  customize prefixes for them

I'm honestly not sure where something like this would live in our config/docs, but I'm also not sure it's necessary (provided we get it right at some point), going a little bit on a tangent, I would honestly remove the customization of these until requested, we can of course keep it parameterized though, so that it's easier to customize the time for that eventually comes.

> Another solution would be to move these toggles to vim variables of the form g:LanguageClient_settings_*.

I mean, maybe we'll end up going to something like server specific setting on a dictionary under some sort of 
```
let g:LanguageClient_extensions = {}
let g:LanguageClient_extensions['rust-analyzer'] = { 'chainingHintsPrefix': ' - ' }
```
But I don't think there's a justification for that now, as the toggles we can control from the settings sent to the server (if the user enables them then it should be clear enough for us that they want us to display them), and the prefixes and separator I'm not convinced need customization right now.

martskins:
> I propose the following alternative scheme:
> ChainingHints - simply Iterator<u32>
> TypeHints - : u32
> ParameterHints - 𝑓(a, b, c)
> 
> What do you think?

I think that would be better yeah. Will need to try it out myself to see if it's clearer (to me of course) but it does seem sensible at first sight.

doc/LanguageClient.txt
Logarithmus: @martskins OK, I went to https://rust-analyzer.github.io/manual.html and saw options for `rust-analyzer.inlayHint.chainingHints` and other hint kinds as well. So it's unreasonable to document them in our project.

All hint kinds in `rust-analyzer` are turned on by default. I know that you think hints are too intrusive to show them by default, but in order to provide expected and consistent behaviour we should enable them by default too. Plus `coc-rust-analyzer`, which seems to be the most popular solution for adding `rust-analyzer` support to `nvim`, does this as well.

Summing up, I'll implement this alternative scheme https://github.com/autozimu/LanguageClient-neovim/pull/1221#issuecomment-835931022, enable hints by default (if not overriden by `settings.json`), remove customization for prefixes and remove the documentation as well.

Logarithmus:
@martskins I've performed the changes from https://github.com/autozimu/LanguageClient-neovim/pull/1221#discussion_r629232168

martskins:
So I've given this a try, there's a couple things that I noticed. 

The first one, is that if you have nested function calls, then the parameter hints get confusing. See line 38 in the first screenshot (from vim) here. I'm not sure how I would display this, maybe something like `f(reader: f(inner))` ? Not entirely convinced, but something like that.

The other thing is that this seems to be hitting a limit in the number of virtual texts. See line 46 in the first screenshot, that same line has a hint in the second screenshot (from VSCode).

Lastly, I think it would be nice to prefix the type parameters with the name of the variable, much like coc.nvim does. So instead of showing `: Sender<Call>, Receiver<Call>`, it would be much nicer to show something like this `tx: Sender<Call>, rx: Receiver<Call>`. I suppose this is something we can get from rust-analyzer too, but I haven't checked.

On your comment about displaying them by default, I agree. If rust-analyzer enables them by default then it makes perfect sense for us to do the same.

Again thank you for working on this, much appreciated.

Screenshot_2021-05-13_09-22-00
Screenshot_2021-05-13_09-22-22

Logarithmus:
@martskins I'm not planning to work on this PR anymore, because neovim 0.5.0 was finally released and someday (when finally have free time) I'll migrate my config to its built-in language server support & `https://github.com/hrsh7th/nvim-compe`. If anyone interested in continuing this PR, feel free to open your own one based on mine.

PR #1220, opened 2021-05-07
Unpin toolchain version

Logarithmus:
Closes https://github.com/autozimu/LanguageClient-neovim/issues/1219

PR #1237, opened 2021-10-28
Update rust toolchain to 1.56.0

gautamg795:
Currently, the selected rust toolchain does not have support for aarch64-apple-darwin (i.e. M1 Mac) and this plugin will not build.
Updating this to the latest stable release seems to build just fine on M1 Macs, and shouldn't cause any negative changes.


gautamg795:
Hm, looks like this was "fixed" on the dev branch already by unpinning the toolchain entirely in #1220. That's probably not the right approach, but anyhow making that decision wasn't the goal of this PR

